PR #47, opened 2017-08-16
Uses cdp v1.1.0

zshaheen:


zshaheen:
Travis is gonna fail until we actually release v1.1.0 of cdp.
That will be next week after some fixes.

PR #65, opened 2017-08-29
Minor plot tweaks

golaz:


zshaheen:
I actually tested this and it works.

PR #79, opened 2017-09-07
Documentation is now on branch 'gh-pages'

golaz:
ACME Diags documentation is now maintained with Sphinx and kept on its own 
separate branch ('gh-pages') so that it can be served directly from github.io.


zshaheen:
This PR deletes some file in `docs` which are used in the viewer.
I'll merge after we release, since these are changes that will require some testing.

golaz:
@zshaheen : I did not realize some files were needed by the viewer; just put the needed files back in `docs`. But we should not have a release with documentation in multiple places.

zshaheen:
@golaz Hmmm, okay. I'll take care of that.

zshaheen:
All of this is already done on master.

PR #83, opened 2017-09-09
Golaz/misc fixes

golaz:
A number of small improvements and fixes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# vi swap files
+*.swp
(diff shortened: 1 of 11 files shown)


golaz:
@zshaheen : this PR includes code to control plot size with mpl by specifying figure size (inches) and resolution (dpi).

```
self.figsize = [8.5, 11.0]
self.dpi = 150
```
The default that I selected will give png figures with size very similar (but not identical) to vcs. If you want, there is an opportunity to better unify vcs and mpl here by redefining how you set the vcs canvas size:

```
canvas_size_w = figsize[0]*dpi
canvas_size_h = figsize[1]*dpi
```

The best would be to do this in the plotting functions where canvas_size is needed and remove the canvas_size_{w,h} parameters.

PR #109, opened 2017-11-09
Issue92 subtitles

chengzhuzhang:
1. added a new parameter 'short_test_name' for user defined model name, todo: update documentation to reflect this new parameter change.
2. added 'year_averaged' to print out as part of the subtitle.


zshaheen:
Looks good, it works with both backends:

vcs:
screen shot 2017-11-09 at 4 21 23 pm

mpl:
screen shot 2017-11-09 at 4 18 31 pm

PR #139, opened 2018-03-23
Model v model

chengzhuzhang:
Add stronger support to model vs model run type
- updates all model_vs_model configuration files 
- when run_type = 'model vs model'. All default configuration file with names *_model_vs_model.cfg are used 
-better presentation for variable names shown in all sets of diags for model_vs_model case


chengzhuzhang:
For Taylor diagram generated for model_vs_model case: reference data is for the reference model, not observational datasets. If there is a need to change, we will modify in a later release

PR #169, opened 2018-06-15
Save figures for individual panel plots

golaz:
Implements new option to save figures for individual panel plots.

This feature is triggered using a new parameter 'output_format_subplot':

`output_format_subplot = []`
Default behavior, no individual plot files saved.

`output_format_subplot = ['png', 'pdf']`
Individual panel plots saved in png and pdf formats.

Supported plot sets:
- lat_lon
- polar
- zonal_mean_2d
- zonal_mean_xy
- cosp_histogram

Only supported with Matplotlib (mpl) backend. Implementation with VCS left for someone else.


zshaheen:
I tested this and it looks great. Thank you Chris.

PR #189, opened 2018-10-31
Analysis scripts more fix

chengzhuzhang:
More fix/modification for processing script for data to be used as time-series input to e3sm_diags. Note that the default cfg files needs to be modified accordingly to reflect these changes, i.e. GPCP and Hadlsst data


chengzhuzhang:
@zshaheen Yes. The one with the bad time-bounds is HadlSST data. After using cdms2 to do sea ice masking, The time bounds was fixed simultaneously.

PR #195, opened 2018-12-12
Climo

zshaheen:
* Allows for time-series files to be used as input, as well as climo files.
* [See here for an example](https://github.com/E3SM-Project/acme_diags/blob/master/examples/model_ts-vs-model_ts/model_ts-vs-model_ts.ipynb).


chengzhuzhang:
That's a lot of commits :)

I tested for several use cases and seems it was pretty stable. I need to test one more thing, if the code base is backward compatible to the newly organized analysis data sets.

zshaheen:
Thank you, @chengzhuzhang and @golaz, for testing this.

PR #220, opened 2019-05-28
Py3 support

zshaheen:
* Python 3 is now the only version of Python we support.


zshaheen:
@chengzhuzhang has been testing this and it looks good on her end.

Issue #223, opened 2019-06-03
500 hPa geopotential height in model_vs_model mode


golaz:
In 'model_vs_model' mode, 500 hPa geopotential field for reference model shows unexpected features (related to topography). Middle panel below:

![20190509 A_WCYCL1950S_CMIP6_LRtunedHR ne30_oECv3_ICG anvil-Z3-500-ANN-global](https://user-images.githubusercontent.com/15743046/58833426-da7b8f80-8605-11e9-8413-deac9631b9f5.png)

The same field plotted in 'model_vs_obs' mode looks as expected. Top panel below:

![ERA-Interim-Z3-500-ANN-global](https://user-images.githubusercontent.com/15743046/58833441-e6675180-8605-11e9-9e91-c3ea4b05186e.png)

Using a recent version of E3SM Diags: 81d0d64fc. Here are the configuration files that were used:

[all_sets_v1-HR_vs_v1-LRtunedHR.py](https://github.com/E3SM-Project/e3sm_diags/files/3249436/all_sets_v1-HR_vs_v1-LRtunedHR.py.txt)

[all_sets_v1-LRtunedHR.py](https://github.com/E3SM-Project/e3sm_diags/files/3249437/all_sets_v1-LRtunedHR.py.txt)

chengzhuzhang:
@golaz Thank you for catching this. I can reproduce this problem with master. And when I shift test/ref data in model vs model run type, the middle panel plot also got the fussy looking. The problem is not shown for model vs obs, cause obs data often available on pressure level, no vertical remapping is needed. Something is wrong for vertical interpolation for reference data (middle panel). I need to look into it.

chengzhuzhang:
A bug when reading in data is fixed to solve this issue. 
Result before fixing:
![iss223](https://user-images.githubusercontent.com/13056557/59223968-140e4680-8b82-11e9-9a73-68eedf4f110a.png)

Result after fixing:
![iss223_fixed](https://user-images.githubusercontent.com/13056557/59223988-1ec8db80-8b82-11e9-8605-195d6fe3b1ea.png)
 
Only reference plot for variables in model vs model run type when vertical interpolation is applied is affected.

chengzhuzhang:
@golaz Thank you for finding this bug! Please close the issue if the fix looks fine...

PR #227, opened 2019-06-10
fixed a bug for issue 223

chengzhuzhang:
Fixed for both lat_lon and polar plots https://github.com/E3SM-Project/e3sm_diags/issues/223. Only for variables in model vs model run type when vertical interpolation is applied is affected.


chengzhuzhang:
Just posted before and after output here: https://github.com/E3SM-Project/e3sm_diags/issues/223

chengzhuzhang:
@zshaheen Thank you for reviewing! Merged.

PR #252, opened 2019-10-04
Update system test

chengzhuzhang:
Update system tests by new way of running and area mean time series test.


chengzhuzhang:
@forsyth2 I just updated the system tests. This is a small change and won't effect the code. Just to  let you know that the system test will be back running with this PR:
```
python all_sets.py -d all_sets.cfg
```

acme_diags/run.py
```diff
@@ -188,8 +188,8 @@ def _remove_attrs_with_default_values(self, param):
 
     def _find_attrs_with_none_default_values(self, param):
         """
-        In the param, remove any parameters that
-        have their default value.
+        In the param, find any parameters that
+        have none default value.
```
forsyth2: I'd say `default value of None` to match the casing of Python's `None`.

forsyth2: I'd say `default value of None` just to match Python's casing of `None`.

chengzhuzhang: Thanks for reviewing. It appears to be a misspelling. I updated the PR.

PR #253, opened 2019-10-14
More cmip support

chengzhuzhang:
The purpose of this PR is to use e3sm_diags as a tool to generate another layer of verification for completeness of v1 cmorized data:
-e3sm_diags now can take xmls files as input.
-variables from land and ocean components were added into derived variable list.

Example output files for cmorized files:https://acme-viewer.llnl.gov/zhang40/cmip_20191014/
In the example folder: all  experiments are grouped into: hist_r1-r5, amip_r1-r3, piControl, CO2(1pct and abrupt4xco2)

Caveat:
-only 2d variables are supported for now.

### DIFF
--- a/acme_diags/derivations/acme.py
+++ b/acme_diags/derivations/acme.py
@@ -615,7 +615,7 @@ derived_variables = {
         (('CLDTOT',), lambda cldtot: convert_units(cldtot, target_units="%"))
     ]),
     'CLOUD': OrderedDict([
-        (('clt',), rename),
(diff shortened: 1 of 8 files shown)


chengzhuzhang:
@golaz I just wanted to let you know this new capability of e3sm_diags to visualize time series of most cmorzied variables. I think you may be interested to look at time series of all the variables form atm/land/ocean components. I think no need to test it out, but let me know if you have any suggestions.

PR #260, opened 2019-11-13
Reduce redundancy in quick guides docs

forsyth2:
Quick guide docs were duplicated for each machine. This pull request reduces this redundancy by adding a Python script that can auto-generate the machine-specific quick guides.


source/quickguides/quick-guide-generic.rst
```diff
@@ -0,0 +1,248 @@
...
+        prefix = '#expand web_address#doc_examples/'
+        param.results_dir = os.path.join(prefix, 'area_mean_with_obs')
+        #param.multiprocessing = True
+        #param.num_workers =  40
+        
+        # We're passing in this new object as well, in
+        # addition to the CoreParameter object.
+        
+        ts_param = AreaMeanTimeSeriesParameter()
+        #ts_param.ref_names = ['none']   # This setting plot model data only
```
forsyth2: What does "This setting plot model data only" mean?

- "Only use this setting if you're plotting model data"?
- "Using this setting will plot only the model data, not the observation data"?
- Something else?

source/quickguides/quick-guide-generic.rst
```diff
@@ -0,0 +1,248 @@
...
+
+Advanced: Running custom diagnostics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The following steps are for 'advanced' users, who want to run custom diagnostics.
+So, most users will not run the software like this.
+
+
+By default, with ``e3sm_diags``,
+a built in set of variables are defined for each diagonostics sets.
+To do a short run, e.g. only running through a subset of variables,
```
forsyth2: Is running through a subset of variables the only way a run would be short? If so, we should use `i.e.` instead of `e.g.`

source/quickguides/quick-guide-generic.rst
```diff
@@ -0,0 +1,248 @@
...
+        variables = ["SST"]
+        ref_name = "HadISST"
+        reference_name = "HadISST/OI.v2"
+        seasons = ["ANN", "DJF", "MAM", "JJA", "SON"]
+        contour_levels = [-1, 0, 1, 3, 6, 9, 12, 15, 18, 20, 22, 24, 26, 28, 29]
+        diff_levels = [-5, -4, -3, -2, -1, -0.5, -0.2, 0.2, 0.5, 1, 2, 3, 4, 5]
+
+
+Run E3SM diagnostics with the ``-d`` parameter.
+Use the :ref:`run script linked here <cori-params-v2>`. And run as following:
```
forsyth2: This link goes to `e3sm_diags/docs/html/quickguides/quick-guide-general.html#cori-params-v2`, which is actually `e3sm_diags/docs/html/quickguides/quick-guide-general.html#running-the-annual-mean-latitude-longitude-contour-set`.

What do we actually want to link to?

source/quickguides/index.rst
```diff
@@ -8,5 +8,5 @@ Quick Guides
    quick-guide-general
```
forsyth2: Do we even want the general quick guide? To show proper code, we need to know what machine we're on. We could just have the three machine-specific quick guides and then have a general guide to list `obs_path`, `test_data_path`, `html_path`, and `web_address` for the remaining machines.

source/quickguides/generate_quick_guides.py
```diff
@@ -0,0 +1,76 @@
...
+                        for mpf_line in mpf:
+                            specific_quick_guide_files[machine_name].write(mpf_line)
+                else:
+                    specific_quick_guide_files[machine_name].write(new_line)
+    for machine_name in machine_names:
+        specific_quick_guide_files[machine_name].close()
+
+
+if __name__ == '__main__':
+    generate_quick_guides()
```
forsyth2: Add newline

source/quickguides/index.rst
golaz: I agree; we don't really need a general quick guide anymore. If a user wants to run on a different machine, they can probably extrapolate from one of the existing guide.

source/quickguides/quick-guide-generic.rst
```diff
@@ -0,0 +1,248 @@
...
+
+Running all the core diagnostics sets
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Core diagnostics set includes:
+**lat_lon**, **zonal_mean_xy**, **zonal_mean_2d**, **polar**, **cosp_histogram**,
+**meridional_mean_2d**.
+These diags share a common parameter space (core parameters).
+To run all these sets without defining set-specific parameters
+(e.g. **plev** for **zonal_mean_2d** and **meridional_mean_2d**.),
```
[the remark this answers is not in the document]
golaz: As a non-native English speaker, I'll defer to you as the proper usage here.

source/quickguides/quick-guide-generic.rst
golaz: Same as above.

golaz:
@forsyth2 : I like your implementation, nice job.

Can you modify the Makefile so that it automatically invokes the python script when needed?

Just a comment on your python script (no need to modify anything). You probably could have avoided building your own search and replace tool by using jinja templates instead. I use them a lot to customize various scripts that I use to process mode output. They are pretty cool and powerful.

source/quickguides/quick-guide-generic.rst
chengzhuzhang: "Using this setting will plot only the model data, not the observation data" is cleaner

source/quickguides/quick-guide-generic.rst
chengzhuzhang: source/quickguides/quick-guide-cori.rst has an anchor point `.. _cori-params-v2:` where it should refer to

source/quickguides/quick-guide-generic.rst
```diff
@@ -0,0 +1,248 @@
...
+        runner.run_diags([param])
+
+
+Run in serial with:
+
+    ::
+
+        python run_e3sm_diags.py
+
+The above run has the same results as :ref:`the parameter file linked here <cori-params-v1>`,
```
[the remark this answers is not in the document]
chengzhuzhang: source/quickguides/quick-guide-cori.rst has an anchor point .. _cori-params-v2: where it should refer to

source/quickguides/generate_quick_guides.py
```diff
@@ -0,0 +1,76 @@
...
+import re
+import subprocess
+
+EXPANSIONS = {
+    'acme1': {
+        'machine_name': 'Acme1',
+        'activation_path': '/usr/local/e3sm_unified/envs/load_latest_e3sm_unified.sh',
+        'obs_path': '/p/user_pub/e3sm/e3sm_diags_data/obs_for_e3sm_diags/',
+        'test_data_path': '/p/user_pub/e3sm/e3sm_diags_data/test_model_data_for_acme_diags/',
+        'html_path': '/var/www/acme/acme-diags/<username>/',
```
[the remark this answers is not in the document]
chengzhuzhang: I think `html_path` is acceptable in this case...

source/quickguides/quick-guide-generic.rst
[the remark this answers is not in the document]
forsyth2: So, each quick-guide should have an anchor `..<machine_name>-params-v2:`? We shouldn't always refer to `cori`, right?

source/quickguides/quick-guide-generic.rst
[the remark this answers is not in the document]
forsyth2: It sounds like there are other set-specific parameters, so `e.g.` makes sense.

source/quickguides/quick-guide-generic.rst
forsyth2: It sounds like there are other ways for a run to be short, so `e.g.` makes sense.

source/quickguides/quick-guide-generic.rst
forsyth2: Same comment as above.

forsyth2:
@golaz Thanks! 

Yes, modifying the Makefile sounds like a good idea. 

I was unaware of Jinja templates - perhaps in the future this would be a good thing to use for the docs. I'm not sure if the templates will work with the `rst` files though; it looks like they are used with `html` files.

source/quickguides/quick-guide-generic.rst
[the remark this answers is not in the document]
chengzhuzhang: I have a second thought, maybe we link to source/quickguides/old/quick-guide-cori.rst and put the anchor point there? Let's only point to `cori`. I don't think we need to do different machines, cause this is only for reference.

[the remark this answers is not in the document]
forsyth2: I'm thinking not even linking to anything. Maybe just display the code block for the parameters file directly. I think it would be confusing to direct to v1 docs, and furthermore to a specific machine in v1 docs. 

Also this links to `all_sets.py` when it seems like the correct params file would be the `myparams.py` earlier in the cori v1 docs (https://e3sm-project.github.io/e3sm_diags/docs/html/quickguides/old/quick-guide-cori.html)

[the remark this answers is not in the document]
chengzhuzhang: I agree that it's a better solution to display the section of codes.

source/quickguides/quick-guide-generic.rst
forsyth2: I think it would be correct to link this to the `lat_lon` run script code block above?

forsyth2:
For this pull request, I will leave the `rst` file generation as a manual step. In future work, we can try to automate that step by updating the Makefile.

Issue #265, opened 2020-01-07
issues and enhancement for parameter plev for zonal mean 2d diags


chengzhuzhang:
Customize plev works when specifying plev as follows:
```
z2d_param = ZonalMean2dParameter()
new_plevs = [1000.0, 900.0, 800.0, 700.0, 600.0, 500.0, 400.0, 300.0, 200.0, 100.0, 70.0, 50.0, 30.0, 20.0, 10.0]
z2d_param.plevs = new_plevs[::-1]
```
It doesn't work setting plevs, using two following methods:
```
z2d_param.plevs = list(numpy.logspace(1.0, 3.0, num=17))
```
or
```
z2d_param.plevs = [10.0, 13.33521432163324, 17.78279410038923, 23.71373705661655, 31.622776601683793, 42.169650342858226, 56.23413251903491, 74.98942093324558, 100.0, 133.3521432163324, 177.82794100389228, 237.13737056616552, 316.22776601683796, 421.6965034285823, 562.341325190349, 749.8942093324558, 1000.0]
```
or
```
z2d_param.plevs =numpy.logspace(1.0, 3.0, num=17)
```
getting errors:
```Traceback (most recent call last):
  File "run_diags_plevs.py", line 26, in <module>
    runner.run_diags([param,z2d_param])
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/acme_diags/run.py", line 26, in run_diags
    final_params = self.get_final_parameters(parameters)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/acme_diags/run.py", line 51, in get_final_parameters
    self._add_parent_attrs_to_children(parameters)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/acme_diags/run.py", line 128, in _add_parent_attrs_to_children
    nondefault_param_child = self._find_attrs_with_nondefault_values(parameters[i])
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.3.0/lib/python3.7/site-packages/acme_diags/run.py", line 202, in _find_attrs_with_nondefault_values
    getattr(new_instance, attr) != getattr(param, attr):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

[...]

PR #266, opened 2020-01-08
Enhance plev for issue #265

chengzhuzhang:
For addressing issue #265
-Revised the set-specific parameter list for zonal-mean-2d and meridional-mean-2d: 
  plevs: needs to be a list
  plot_plevs: if set `True` the plevs ticks set at plevs values. [default to false]
  plot_log_plevs: if set `True` plot log-scale. [default to false]

-Add a check to ensure plevs is increasing monotonically.

### DIFF
--- a/acme_diags/driver/utils/general.py
+++ b/acme_diags/driver/utils/general.py
@@ -11,6 +11,20 @@ import cdms2
 from acme_diags import container
 from acme_diags.derivations.default_regions import regions_specs
 
+def strictly_increasing(L):
+    return all(x<y for x, y in zip(L, L[1:]))
+
+def strictly_decreasing(L):
+    return all(x>y for x, y in zip(L, L[1:]))
+
+def monotonically_decreasing(L):
+    return all(x>=y for x, y in zip(L, L[1:]))
+
+def monotonically_increasing(L):
+    return all(x<=y for x, y in zip(L, L[1:]))
+
+def monotonic(L):
+        return monotonically_increasing(L) or monotonically_decreasing(L)
 
 
 def adjust_time_from_time_bounds(var):
--- a/acme_diags/parameter/meridional_mean_2d_parameter.py
+++ b/acme_diags/parameter/meridional_mean_2d_parameter.py
@@ -1,11 +1,33 @@
 import numpy
 from .core_parameter import CoreParameter
-
+from acme_diags.driver.utils.general import monotonic
 
 class MeridionalMean2dParameter(CoreParameter):
     def __init__(self):
         super(MeridionalMean2dParameter, self).__init__()
-        self.plevs = numpy.logspace(2.0, 3.0, num=17)
+        self.plevs = numpy.logspace(2.0, 3.0, num=17).tolist()
+        self.plot_log_plevs = False
+        self.plot_plevs = False
         # Granulating plevs causes duplicate plots in this case.
-        # So keep all of the default values except plevs. 
+        # So keep all of the default values except plevs.
         self.granulate.remove('plevs')
+
+    def check_values(self):
+        plevs = self.plevs
+        if not isinstance(plevs,list):
+           msg = 'plevs needs to be a list'
+           raise RuntimeError(msg)
+
+        if len(plevs) > 1:
+            if monotonic(plevs):
+                if plevs[0] > plevs[1]:
+                    # Force plevs to be monotonically increasing by reversing the list.
+                    plevs = plevs[::-1]
+                    self.plevs = plevs
+            else:
+                msg = 'plevs should be monotonically increasing or decreasing'
+                raise RuntimeError(msg)
+        else:
+            msg = "At least 2 plevs needed"
+            raise RuntimeError(msg)
+
--- a/acme_diags/parameter/zonal_mean_2d_parameter.py
+++ b/acme_diags/parameter/zonal_mean_2d_parameter.py
@@ -1,11 +1,35 @@
 import numpy
 from .core_parameter import CoreParameter
-
+from acme_diags.driver.utils.general import monotonic
 
 class ZonalMean2dParameter(CoreParameter):
     def __init__(self):
         super(ZonalMean2dParameter, self).__init__()
-        self.plevs = numpy.logspace(2.0, 3.0, num=17)
+        self.plevs = numpy.logspace(2.0, 3.0, num=17).tolist()
+        self.plot_log_plevs = False 
+        self.plot_plevs = False 
         # Granulating plevs causes duplicate plots in this case.
         # So keep all of the default values except plevs.
         self.granulate.remove('plevs')
+
+    def check_values(self):
+        plevs = self.plevs
+        if not isinstance(plevs,list):
+           msg = 'plevs needs to be a list'
+           raise RuntimeError(msg)
+
+
+        if len(plevs) > 1:
+            if monotonic(plevs):
+                if plevs[0] > plevs[1]:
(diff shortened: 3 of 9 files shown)


acme_diags/driver/utils/general.py
```diff
@@ -11,6 +11,20 @@
 from acme_diags import container
 from acme_diags.derivations.default_regions import regions_specs
 
+def strictly_increasing(L):
+    return all(x<y for x, y in zip(L, L[1:]))
+
+def strictly_decreasing(L):
+    return all(x>y for x, y in zip(L, L[1:]))
+
+def non_increasing(L):
```
forsyth2: It might be better to call this function `monotonically_decreasing` and the next one `monotonically_increasing`; that way we aren't naming things in the negative.

acme_diags/parameter/meridional_mean_2d_parameter.py
```diff
@@ -1,11 +1,32 @@
 import numpy
 from .core_parameter import CoreParameter
-
+from acme_diags.driver.utils.general import strictly_increasing,strictly_decreasing,non_increasing,non_decreasing,monotonic
 
 class MeridionalMean2dParameter(CoreParameter):
     def __init__(self):
         super(MeridionalMean2dParameter, self).__init__()
-        self.plevs = numpy.logspace(2.0, 3.0, num=17)
+        self.plevs = numpy.logspace(2.0, 3.0, num=17).tolist()
```
forsyth2: Does changing this to a `list` cause any problems?

acme_diags/parameter/meridional_mean_2d_parameter.py
```diff
@@ -1,11 +1,32 @@
...
+    def check_values(self):
+        plevs = self.plevs
+        if not isinstance(plevs,list):
+           msg = 'plevs needs to be a list'
+           raise RuntimeError(msg)
+
+        if len(plevs) > 1:
+            if monotonic(plevs):
+                if plevs[0] > plevs[1]:
+                    plevs = plevs[::-1]
```
forsyth2: What are you trying to accomplish here? If `plevs[0] > plevs[1]` is `True` then `plevs` must be montonically decreasing. If so, then you reverse the entire list. This would make it monotonically increasing. Why don't you just check that plevs is`montonically_increasing` in the first place then?

acme_diags/parameter/meridional_mean_2d_parameter.py
```diff
@@ -1,11 +1,32 @@
...
+                    self.plevs = plevs
+            else:
+                msg = 'plevs should be monotonicly increasing or decreasing'
```
forsyth2: `monotonically`

acme_diags/parameter/zonal_mean_2d_parameter.py
```diff
@@ -1,11 +1,35 @@
 import numpy
 from .core_parameter import CoreParameter
-
+from acme_diags.driver.utils.general import strictly_increasing,strictly_decreasing,non_increasing,non_decreasing,monotonic
```
forsyth2: Same comments as last file. Actually, these files are so similar that you might consider creating a `Mean2dParameter` class for both to inherit from.

acme_diags/parser/zonal_mean_2d_parser.py
```diff
@@ -20,6 +20,21 @@ def load_default_args(self, files=[]):
             type=float,
             nargs='+',
             dest='plevs',
-            help='Selected pressure level.',
+            help='Selected pressure level.[take list as input]',
```
forsyth2: Might want to make a `Mean2dParser` for both to inherit from.

acme_diags/parameter/zonal_mean_2d_parameter.py
forsyth2: Perhaps that could be a separate GitHub issue; since these files existed before this pull request, it wouldn't make sense to do that refactor in this pull request.

acme_diags/parser/zonal_mean_2d_parser.py
forsyth2: Same comment as above re: GitHub issue to make this common parent class.

acme_diags/run.py
```diff
@@ -199,7 +199,7 @@ def _find_attrs_with_nondefault_values(self, param):
                 continue
             
             if hasattr(new_instance, attr) and \
-                getattr(new_instance, attr) != getattr(param, attr):
+                getattr(new_instance, attr) != getattr(param, attr):   #this only valid when the attri values are lists not numpy array
```
forsyth2: Add space after `#`. `attri` should be `attr`. If this is only valid if `attr` values are lists, what happens in the code if they are not?

tests/system/all_sets.py
```diff
@@ -1,12 +1,15 @@
...
 from acme_diags.run import Run
+import numpy
 
 thing = Run()
 param = CoreParameter()
 ts_param = AreaMeanTimeSeriesParameter()
+z2d_param = ZonalMean2dParameter()
 
-thing.run_diags([param, ts_param])
+thing.run_diags([param, ts_param,z2d_param])
```
forsyth2: No need to test `meridional`?

acme_diags/parameter/meridional_mean_2d_parameter.py
chengzhuzhang: plevs being a numpy array caused the problem, then I converted it list. It doesn't show problem with testing

acme_diags/parameter/meridional_mean_2d_parameter.py
chengzhuzhang: `monotonic` is either decreasing or increasing. I need to make sure the list is monotonic first.

acme_diags/parameter/zonal_mean_2d_parameter.py
chengzhuzhang: True. Let's keep it this way for now.

acme_diags/run.py
chengzhuzhang: i.g., if the `attr` value is a numpy array, the logic expression returns a Boolean array, which caused the problem.

tests/system/all_sets.py
chengzhuzhang: yeah, should add `meridional`

acme_diags/driver/utils/general.py
chengzhuzhang: okay

acme_diags/parameter/meridional_mean_2d_parameter.py
forsyth2: This is true, but then you effectively force `plevs` to be monotonically increasing, so why not just check that? You have the function for it in `utils.general`? Are users using montonically decreasing `plevs` so frequently that we'll just reverse the list for them?

acme_diags/run.py
forsyth2: So, by the time the code reaches this point, we know for sure that `attr` values are lists, not numpy arrays?

acme_diags/parameter/meridional_mean_2d_parameter.py
chengzhuzhang: it's not obvious that what order users will put in for plevs. I would say 50-50 of chance.

acme_diags/run.py
chengzhuzhang: yes. The `check_value` function in the parameter scripts should take care of this.

chengzhuzhang: Another way is to find a Boolean expression valid to check match for both list and numpy array.

PR #272, opened 2020-01-23
add plevs parameters to core

chengzhuzhang:
Add plev related parameters/parsers back to core sets, so that the old configuration files from users are still compatible.


chengzhuzhang:
@golaz with this fix, I have tested that your current configuration file should work (see [here](https://compy-dtn.pnl.gov/zhan429/981-1990_v1/viewer/)) for zonal mean 2d and meridional 2d plots. Please let me know if this is okay to merge?

golaz:
@chengzhuzhang : thanks, this will help with backwards compatibility. We should still strongly encourage users to switch to the v2 API. I'm trying to use it as much as possible from now on. Please go ahead and merge the changes.

PR #274, opened 2020-01-28
Update Cori paths

forsyth2:
Cori file paths have changed. Therefore, the paths in the documentation must be changed. I left the paths in `quickguides/old` the same, for history.


forsyth2:
@chengzhuzhang Ready for review. I ran `python source/quickguides/generate_quick_guides.py` and `make html` as in https://e3sm-project.github.io/e3sm_diags/docs/html/contributing.html#checkout-and-edit-documentation

You probably only need to review changes in `source/`

Issue #280, opened 2020-02-26
Difference plots show small residual running on same datasets


chengzhuzhang:
Reported by @oksanaguba When contour levels are not specified, small residual showing up on difference map. It could be truncation error somewhere...

![-CCN3-ANN-global](https://user-images.githubusercontent.com/13056557/75390645-7b4d6180-589d-11ea-9fbb-8b85baaa53b1.png)

More examples are shown here: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/zhang40/def/viewer/zonal_mean_2d/index.html

golaz:
Are these the exact same fields being plotted?

chengzhuzhang:
Yes, according to Oksana, she's comparing the data model runs as test and reference.

oksanaguba:
Here is my script
```
import os
import numpy
from acme_diags.parameter.core_parameter import CoreParameter
from acme_diags.parameter.zonal_mean_2d_parameter import ZonalMean2dParameter
from acme_diags.run import runner

param = CoreParameter()

#DEFAULT MODEL
param.reference_data_path = '/lcrc/group/acme/onguba/acme_scratch/anvil/theta-ne30-default-feb21/run/climoout.136284-0002-0006'
#OBS
#param.reference_data_path = '/lcrc/soft/climate/e3sm_diags_data/obs_for_e3sm_diags/climatology/'
param.test_data_path = '/home/onguba/as/theta-ne30-default-feb21/run/climoout.136284-0002-0006'
prefix = '/lcrc/group/acme/public_html/diagnostic_output/onguba/theta/theta-ne30-default-feb21.136284.0002-0006'
param.results_dir = os.path.join(prefix,'def')  #'zonal-vs-default-custom-plev-log')
param.test_title = 'theta-ne30-default-feb21.136284'
param.run_type = 'model_vs_model'
#param.run_type = 'model_vs_obs'
param.multiprocessing = True
param.num_workers =  8 # was 8


z2d_param = ZonalMean2dParameter()
#z2d_param.plevs = list(numpy.logspace(1.0, 3.0, num=17))
#print(z2d_param.plevs)
z2d_param.plevs =  [1000.0, 900.0, 800.0, 700.0, 600.0, 500.0, 400.0, 300.0, 200.0, 100.0, 70.0, 50.0, 30.0, 20.0, 10.0]
z2d_param.plot_log_plevs = True
#z2d_param.plevs = new_plevs[::-1]

[...]

PR #281, opened 2020-02-27
Issue280

chengzhuzhang:
This PR is to fix https://github.com/E3SM-Project/e3sm_diags/issues/280.


oksanaguba:
Seems good. If you want i can test it with the same run as before, since i am running diagnostics from a clone, not preinstalled. Thanks for the fix!

chengzhuzhang:
@oksanaguba Thank you for your review! If it possible please let me know if this new branch works to generate the diff plots as expected...

PR #243, opened 2019-08-17
Disable MPI in cdms2.tvariable

xylar:
This should fix an issue where e3sm_diags is not able to deepcopy variables because of (unneeded) MPI comm objects stored in the variable.

closes #242

Issue #242, opened 2019-08-10
e3sm_diags in unified_env missing plots - mpi4py issue


chengzhuzhang:
@xylar I'm testing e3sm_diags on compy. It occurs to me that some diagnostics plots are not generated under unified_env_py3.7, as following:
https://compy-dtn.pnl.gov/zhan429/lat_lon_demo/viewer/latitude-longitude/index.html
(e.g., SOLIN global ceres_ebaf_toa_v2.8)

However when I tried to run e3sm_diags in its own conda environment (1.7.1). All plots are generated, see following:
https://compy-dtn.pnl.gov/zhan429/lat_lon_demo_1.7.1_py3/viewer/latitude-longitude/index.html

The error I have when running from e3sm_unified_py3.7:
```
Error in acme_diags.driver.lat_lon_driver
Traceback (most recent call last):
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/site-packages/acme_diags/acme_diags_driver.py", line 211, in run_diag
    single_result = module.run_diag(parameters)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/site-packages/acme_diags/driver/lat_lon_driver.py", line 168, in run_diag
    region, mv1, mv2, land_frac, ocean_frac, parameter)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/site-packages/acme_diags/driver/utils/general.py", line 143, in select_region
    var1, land_ocean_frac, low_limit=region_value)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/site-packages/acme_diags/driver/utils/general.py", line 199, in mask_by
    var = copy.deepcopy(input_var)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/copy.py", line 161, in deepcopy
    y = copier(memo)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/site-packages/numpy/ma/core.py", line 6045, in __deepcopy__
    copied.__dict__[k] = deepcopy(v, memo)
  File "/compyfs/software/e3sm-unified/base/envs/e3sm_unified_1.2.6/lib/python3.7/copy.py", line 169, in deepcopy
    rv = reductor(4)
[...]

Issue #212, opened 2019-04-05
MPI failure in batch mode on Edison/Cori?


kaabre:
When ran in batch mode, diagnostic runs fail while initiating MPI and do not produce output.  @beharrop ran into this problem on Edison but I've also reproduced it on Cori, under both the latest shifter image and the v1.5.0 image.  The demo also fails this way (lat_lon and all_sets demo).  However, when run on an interactive node with the exact same commands and parameters, it does not fail (at least in the latest version).  

Fri Mar 29 13:38:54 2019: [unset]:_pmi_alps_init:alps_get_placement_info returned with error -1
Fri Mar 29 13:38:54 2019: [unset]:_pmi_init:_pmi_alps_init returned -1
[Fri Mar 29 13:38:54 2019] [c5-0c0s14n3] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(537):
MPID_Init(247).......: channel initialization failed
MPID_Init(636).......:  PMI2 init failed: 1

zshaheen:
Hi @kaabre, I'll try to reproduce this error. Is this the full output? If not, can you post it? Thanks!

chengzhuzhang:
Hello @kaabre, do you still have this issue?

zshaheen:
Others, like Chris and Qi are having this issue. I'm going to look at this.

zshaheen:
I'm also getting issues just running the container, please see below. I'm in contact with NERSC support.

```
$ shifter --volume=/global/u2/z/zshaheen/github/acme_diags/tests/system:/reference_data_path --volume=/global/u2/z/zshaheen/github/acme_diags/tests/system:/test_data_path --volume=/global/u2/z/zshaheen/github/acme_diags/tests/system/all_sets_results:/results_dir --image=docker:e3sm/e3sm_diags:latest -- e3sm_diags                                       
ERROR: unclean exit from bind-mount routine. /var/udiMount/reference_data_path may still be mounted.
BIND MOUNT FAILED from /var/udiMount//global/u2/z/zshaheen/github/acme_diags/tests/system to /var/udiMount/reference_data_path
FAILED to setup user-requested mounts.
FAILED to setup image.
```

tangq:
@zshaheen , good to know that you can reproduce the errors. I will be interested in knowing the solutions.

zshaheen:
Some good news. @dnadeau4 stated that running in the interactive mode works.

PR #287, opened 2020-03-13
Find bad file descriptor imports

forsyth2:
Find import lines that cause the following error:
```
Fatal error in MPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(572)..............: 
MPID_Init(224).....................: channel initialization failed
MPIDI_CH3_Init(105)................: 
MPID_nem_init(324).................: 
MPID_nem_tcp_init(178).............: 
MPID_nem_tcp_get_business_card(425): 
MPID_nem_tcp_init(384).............: gethostbyname failed, ml-9624328 (errno 1)
[unset]: write_line error; fd=-1 buf=:cmd=abort exitcode=3191567
:
system msg for write_line failure : Bad file descriptor
```
We need to find out why these imports cause the above error on my Mac intermittently;  I don't always get this error.

### DIFF
--- a/acme_diags/acme_diags_driver.py
+++ b/acme_diags/acme_diags_driver.py
@@ -20,7 +20,7 @@ import traceback
 import subprocess
 import cdp.cdp_run
 import acme_diags
-import cdms2.tvariable 
+#import cdms2.tvariable
 from acme_diags.parameter.core_parameter import CoreParameter
 from acme_diags.parser import SET_TO_PARSER
 from acme_diags.parser.core_parser import CoreParser
--- a/acme_diags/derivations/default_regions.py
+++ b/acme_diags/derivations/default_regions.py
@@ -1,4 +1,4 @@
-import cdutil
+#import cdutil
 
 regions_specs = {
     'NHEX': {'domain': cdutil.region.domain(latitude=(30., 90, 'ccb'))},
--- a/acme_diags/driver/utils/climo.py
+++ b/acme_diags/driver/utils/climo.py
@@ -1,6 +1,6 @@
 import numpy as np
 import numpy.ma as ma
-import cdms2
+#import cdms2
 
 
 def climo(var, season):
--- a/acme_diags/driver/utils/dataset.py
+++ b/acme_diags/driver/utils/dataset.py
@@ -8,7 +8,7 @@ import glob
 import re
 import collections
 import traceback
-import cdms2
+#import cdms2
 import acme_diags.derivations.acme
 from acme_diags.driver import utils
 from . import climo
--- a/acme_diags/driver/utils/general.py
+++ b/acme_diags/driver/utils/general.py
@@ -4,10 +4,10 @@ import os
 import copy
 import pwd
 import grp
-import cdutil
-import MV2
-import genutil
-import cdms2
+# import cdutil
+# import MV2
+# import genutil
+# import cdms2
 from acme_diags import container
 from acme_diags.derivations.default_regions import regions_specs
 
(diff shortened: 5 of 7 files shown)


forsyth2:
@chengzhuzhang @golaz  It appears the `Bad file descriptor` bug is linked to the imports I've commented out here. I'm not sure why this is so or how to fix the bug (naturally, we can't actually run the code without these imports). Any thoughts?

chengzhuzhang:
@forsyth2 Thank you for tracking down this error. 
https://github.com/E3SM-Project/e3sm_diags/issues/242
https://github.com/E3SM-Project/e3sm_diags/pull/243
See above PRs, which were made to disable mpi to fix the issue that e3sm_diags can't be running in e3sm_unified. It seems like we need to reassess this PR and to find a work-around..

zshaheen:
I remember getting something similar. Running locally on my macOS machine while VPN'ed into LLNL, I would get a similar error, not sure if it's the exact same thing. Everything was fine when not VPN'ed in. You might want to try that and see if that's the issue. This was due to something within CDMS.

Also #212 might be a similar issue.

forsyth2:
@zshaheen Thanks! Yes, disconnecting from VPN does in fact fix the issue. 

@chengzhuzhang We should still investigate why this is the case. It looks like more imports than CDMS caused issues, but maybe CDMS was the root cause?

chengzhuzhang:
Thank you! @zshaheen 
@forsyth2, yeah, do you know what other imports could cause this issue. If it is a CDMS issue, we can fix it..

zshaheen:
@forsyth2, I remember that cdms was the root cause. Even if you comment out all `import cdms2` statements in `e3sm_diags`, you'll probably still get the error since other CDAT packages import cdms.

When investigating this with @doutriaux1 and @dnadeau4 some time (1+ years) ago, it was determined that finding out why this happens and fixing it wasn't worth the time. Most people at the time didn't run `e3sm_diags` on macOS and it was only a problem for the developers who were VPN'ed in.

If you still want to, you'll need to dig into cdms. Good luck.

chengzhuzhang:
@zshaheen Thank you for chiming in! Ryan is also on CDAT team seeking to revamp CDAT with new structure and dependencies, xarray and dask..etc. It's probably not worth to fix in current version of CDAT. But it might be a good learning progress for Ryan to get familiar CDAT and try avoid the similar issues in the long run. Good to hear from you!

doutriaux1:
Might be my memory giving up, especially since I'm no longer on the project but I seem to recall something similar  being due to vpn and macs. Are you under VPN?

forsyth2:
@chengzhuzhang I'm closing this pull request since 1) the commented imports should definitely not be merged and 2) the actual code changes we need would be in CDMS, not in e3sm_diags.

Should I make a GitHub issue in CDMS to at least track this? I didn't see it listed in https://github.com/CDAT/cdms/issues.

chengzhuzhang:
@forsyth2 Can we keep this issue open, so that we can keep track and if other users encounter similar issues they will have a reference? It is a good idea to file a CDMS issue too.

forsyth2:
@chengzhuzhang I created a corresponding issue in CDMS (https://github.com/CDAT/cdms/issues/391) and referenced this pull request. I don't think it makes sense to reopen this pull request though -- I can make a copy of the CDMS issue in this repo, perhaps?

chengzhuzhang:
@forsyth2 oh, I thought this is an issue... Yes, we should have an issue instead of a PR.

forsyth2:
@chengzhuzhang Yeah, I would have made an issue for this but I needed a pull request to show the import lines that caused problems. I just created #287 in this repo to link to the one in CDMS.

PR #293, opened 2020-03-24
add script processing ERA5

chengzhuzhang:
Started to add ERA5 variables to our obs data sets. This PR is specifically for ERA5 U wind component, which has denser plevs above 100mb available than current ERA-interim data set, in preparation for QBO diags.


chengzhuzhang:
This was covered by PR#308

PR #295, opened 2020-03-24
ENSO updates

### LINKED ISSUE #292: ENSO diags update
For the ENSO diags:
- Users should be able to specify year range for test and reference individually. while allow to use start_year and end_year if  they are the same for both sets of data.
- Use sst = data.get_timeseries_variable('TS') instead of ('SST') for model data, because 'TS' is usually more available
-We need a new entry for ‘LHFLX’ in derived variable list, so that when  ‘QFLX’ is present, the diags will still run.

forsyth2:
Resolves #292.

- Update derivations for `NET_FLUX_SRF` and `LHFLX`.
- Change code to use `TS` if `SST` is not available.
- Update `start_yr` and `end_yr` parameter handling.

### DIFF
--- a/acme_diags/derivations/acme.py
+++ b/acme_diags/derivations/acme.py
@@ -97,6 +97,14 @@ def qflxconvert_units(var):
         var.units = 'mm/day'
     return var
 
+
+def qflx_convert_to_lhflx(qflx):
+    # QFLX units: kg/((m^2)*s)
+    # Multiply by the latent heat of condensation/vaporization (in J/kg)
+    # kg/((m^2)*s) * J/kg = J/((m^2)*s) = (W*s)/((m^2)*s) = W/(m^2)
+    return qflx * 2.5e6
+
+
 def pminuse_convert_units(var):
     if var.units == 'kg/m2/s' or var.units == 'kg m-2 s-1' or var.units == 'kg/s/m^2':
         # need to find a solution for units not included in udunits
@@ -387,6 +395,8 @@ derived_variables = {
         (('prsn',), lambda prsn: qflxconvert_units(rename(prsn))),
         (('PRECSC', 'PRECSL'), lambda precsc, precsl: precst(precsc, precsl))
     ]),
+    # Sea Surface Temperature: Degrees C
+    # Temperature of the water, not the air. Ignore land.
     'SST': OrderedDict([
         # lambda sst: convert_units(rename(sst),target_units="degC")),
         (('sst',), rename),
@@ -517,8 +527,10 @@ derived_variables = {
         (('rsdscs',), rename),
         (('rsdsc',), rename)
     ]),
+    # Net surface heat flux: W/(m^2)
     'NET_FLUX_SRF': OrderedDict([
         (('FSNS','FLNS','LHFLX','SHFLX'), lambda fsns, flns, lhflx, shflx: netflux4(fsns, flns, lhflx, shflx)),
+        (('FSNS','FLNS','QFLX','SHFLX'), lambda fsns, flns, qflx, shflx: netflux4(fsns, flns, qflx_convert_to_lhflx(qflx), shflx)),
         (('rsds','rsus','rlds','rlus','hfls','hfss'), lambda rsds, rsus, rlds, rlus, hfls, hfss: netflux6(rsds, rsus, rlds, rlus, hfls, hfss))
     ]),
     'FLUT': OrderedDict([
@@ -599,12 +611,15 @@ derived_variables = {
         (('TREFHT_LAND',), lambda t: convert_units(t, target_units="DegC")),
         (('tas',), lambda t: convert_units(t, target_units="DegC"))
     ]),
+    # Surface water flux: kg/((m^2)*s)
     'QFLX': OrderedDict([
         (('evspsbl',), rename),
         (('QFLX',), lambda qflx: qflxconvert_units(qflx))
     ]),
+    # Surface latent heat flux: W/(m^2)
     'LHFLX': OrderedDict([
-        (('hfls',), rename)
+        (('hfls',), rename),
+        (('QFLX',), lambda qflx: qflx_convert_to_lhflx(qflx))
     ]),
     'SHFLX': OrderedDict([
         (('hfss',), rename)
@@ -777,6 +792,8 @@ derived_variables = {
     'AODABS': OrderedDict([
         (('abs550aer',), rename)
     ]),
+    # Surface temperature: Degrees C
+    # (Temperature of the surface (land/water) itself, not the air)
     'TS': OrderedDict([
         (('ts',), rename)
     ]),
--- a/acme_diags/driver/enso_diags_driver.py
+++ b/acme_diags/driver/enso_diags_driver.py
@@ -14,7 +14,7 @@ from acme_diags.metrics import rmse, corr, min_cdms, max_cdms, mean, std
 from acme_diags.plot.cartopy.enso_diags_plot import plot_map, plot_scatter
 
 
-def calculate_nino_index(nino_region_str, parameter):
+def calculate_nino_index(nino_region_str, parameter, test=False, ref=False):
     """
     Use the built-in HadISST nino index time series from http://www.esrl.noaa.gov/psd/gcos_wgsp/Timeseries/
     for observation datasets and when model output is not available.
@@ -24,8 +24,15 @@ def calculate_nino_index(nino_region_str, parameter):
     data_file = ''.join(['enso_', nino_region_str, '.long.data'])
     nino_index_path = os.path.join(acme_diags.INSTALL_PATH, 'enso_diags', data_file)
     sst_orig = numpy.loadtxt(nino_index_path, skiprows=1, max_rows=149)  # load data up to year 2018 from 1870
-    start = int(parameter.start_yr)
-    end = int(parameter.end_yr)
+    if test:
+        start = int(parameter.test_start_yr)
+        end = int(parameter.test_end_yr)
+    elif ref:
+        start = int(parameter.ref_start_yr)
+        end = int(parameter.ref_end_yr)
+    else:
+        start = int(parameter.start_yr)
+        end = int(parameter.end_yr)
     sst_years = sst_orig[:,0].astype(int)
     try:
         start_ind = numpy.where(sst_years == start)[0][0]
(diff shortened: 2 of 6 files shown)


acme_diags/driver/enso_diags_driver.py
```diff
@@ -54,7 +54,7 @@ def calculate_nino_index_model(data, nino_region_str, parameter):
     """
 
     try:
-        sst = data.get_timeseries_variable('SST')
+        sst = data.get_timeseries_variable('TS')
```
forsyth2: Is this change ok? If there are any islands in the `nino_region` wouldn't `TS` then also include land surface temperatures along with the sea surface temperatures?

run.py
```diff
@@ -0,0 +1,29 @@
+# TODO: Remove before merging
```
forsyth2: Remove this file before merging

acme_diags/derivations/acme.py
```diff
@@ -604,7 +604,8 @@ def cosp_histogram_standardize(cld):
         (('QFLX',), lambda qflx: qflxconvert_units(qflx))
     ]),
     'LHFLX': OrderedDict([
-        (('hfls',), rename)
+        (('hfls',), rename),
+        (('QFLX'), lambda qflx: qflxconvert_units(qflx))
```
forsyth2: Is this the appropriate derivation?

acme_diags/parameter/enso_diags_parameter.py
```diff
@@ -11,6 +11,18 @@ def __init__(self):
         self.ref_timeseries_input = True
         self.test_timeseries_input = True
 
+        if hasattr(self, 'start_yr'):
```
forsyth2: If `test_start_yr` is set, use that value. Else, use `start_yr`. Similar logic for the other three parameters.

acme_diags/driver/enso_diags_driver.py
chengzhuzhang: This change is okay. Good point about the island, I visually checked all the `nino_regions`, I think use `TS` in this case is a pretty safe assumption.

acme_diags/derivations/acme.py
chengzhuzhang: Yes! By searching through the derived variable list: `NET_FLUX_SRF`, will also impacted by lacking of `LHFLX`, Can you please and an entry for `NET_FLUX_SRF` with `QFLX`.

acme_diags/parameter/enso_diags_parameter.py
chengzhuzhang: The below lines in `check_value` function need to be updated accordingly.
```     
   if not (hasattr(self, 'start_yr') and hasattr(self, 'end_yr')):
            msg = "You need to define both the 'start_yr' and 'end_yr' parameter."
            raise RuntimeError(msg)
```

acme_diags/derivations/acme.py
forsyth2: Is that correct? For instance, `FLDSC` on https://github.com/E3SM-Project/e3sm_diags/blob/master/acme_diags/derivations/acme.py#L495 can be derived from `FLNSC`, which itself can be derived from `sfc_net_lw_clr_mon`, yet we do not list this for `FLDSC` also. 

Are you saying `NET_FLUX_SRF` should have a line where the `LHFLX` in https://github.com/E3SM-Project/e3sm_diags/blob/master/acme_diags/derivations/acme.py#L521 is replaced with `QFLX`? That seems redundant.

acme_diags/parameter/enso_diags_parameter.py
forsyth2: What exactly needs to be checked? If one is defined, the other still needs to be. 

If `start_yr` isn't defined but `test_start_yr` and `ref_start_yr` are equal, then I suppose we could also set `start_yr` to be that value.

acme_diags/derivations/acme.py
chengzhuzhang: Not replaced, but in addition, to add a line using QFLX instead of LHFLX to cover the case only LHFLX is present

acme_diags/derivations/acme.py
```diff
@@ -519,7 +519,8 @@ def cosp_histogram_standardize(cld):
     ]),
     'NET_FLUX_SRF': OrderedDict([
         (('FSNS','FLNS','LHFLX','SHFLX'), lambda fsns, flns, lhflx, shflx: netflux4(fsns, flns, lhflx, shflx)),
-        (('rsds','rsus','rlds','rlus','hfls','hfss'), lambda rsds, rsus, rlds, rlus, hfls, hfss: netflux6(rsds, rsus, rlds, rlus, hfls, hfss))
+        (('rsds','rsus','rlds','rlus','hfls','hfss'), lambda rsds, rsus, rlds, rlus, hfls, hfss: netflux6(rsds, rsus, rlds, rlus, hfls, hfss)),
+        (('QFLX'), lambda qflx: qflxconvert_units(qflx))
```
forsyth2: @chengzhuzhang Let me know what the correct conversion function should be.

acme_diags/driver/enso_diags_driver.py
```diff
@@ -54,15 +54,26 @@ def calculate_nino_index_model(data, nino_region_str, parameter):
     """
 
     try:
-        sst = data.get_timeseries_variable('SST')
+        try:
+            # Try sea surface temperature first.
+            sst = data.get_timeseries_variable('SST')
```
forsyth2: @chengzhuzhang I thought it would be good to at least try `SST` first and otherwise default to `TS`. Let me know if we should always go with `TS`.

test.cfg
```diff
@@ -0,0 +1,13 @@
+[#]
```
forsyth2: Remove this file before merging.

acme_diags/driver/enso_diags_driver.py
```diff
@@ -54,15 +54,26 @@ def calculate_nino_index_model(data, nino_region_str, parameter):
...
+        except RuntimeError as e:
+            if str(e).startswith('Neither does SST nor the variables in'):
+                print(e)
```
forsyth2: In the two `print(e)` lines it might be good to do `print('Handling exception=", e)` instead, to make it clear the user doesn't need to be overly concerned.

tests/system/all_sets.cfg
```diff
@@ -168,11 +168,30 @@ print_statements  = True
...
-plot_type = "scatter"
-case_id = "TREFHT"
+case_id = 'TREFHT-response-map-start-yrs'
 debug = True
+diff_colormap = "BrBG"
+output_format_subplot = ["png"]
+regions = ["20S20N"]
 results_dir = "all_sets_results_test"
+seasons = ["ANN"]
+test_start_yr='2012'
```
forsyth2: @chengzhuzhang Here we test using the following parameter combination: `test_start_yr`, `ref_start_yr`, and `end_yr`. Should we test others or do you think this is sufficient?

acme_diags/derivations/acme.py
chengzhuzhang: `QFLX` is surface water flux with unit ["kg/m2/s"]
`LHFLX`is surface latent heat flux with unit ["W/m2"]
LHFLX = QFLX * 2.5 x 10^6     where 2.5x10^6 is latent heat of condensation/vaporization[J/kg]

acme_diags/driver/enso_diags_driver.py
chengzhuzhang: yeah, good idea.

tests/system/all_sets.cfg
chengzhuzhang: I think most use will be having test_start_yr test_end_yr ref_start_yr ref_end_yr. Maybe we should have this set test out instead?

acme_diags/derivations/acme.py
```diff
@@ -387,6 +387,8 @@ def cosp_histogram_standardize(cld):
         (('prsn',), lambda prsn: qflxconvert_units(rename(prsn))),
         (('PRECSC', 'PRECSL'), lambda precsc, precsl: precst(precsc, precsl))
     ]),
+    # Sea Surface Temperature: Degrees C
```
forsyth2: @chengzhuzhang I added comments to some variables. If you feel strongly that these comments should not be here, I can take them out. That said, I think they really help aid understanding of what's going on in this file. (Rather than commenting on every variable in this file, I figure I can add comments as needed in this and future pull requests).

acme_diags/derivations/acme.py
forsyth2: @chengzhuzhang I added this formula to the `LHFLX` derivation below. I don't see how it makes sense units-wise to put `QFLX` here at all.

tests/system/all_sets.cfg
```diff
@@ -168,11 +168,51 @@ print_statements  = True
...
+[#]
+sets = ["enso_diags"]
+case_id = 'TREFHT-response-map-test-ref-yrs'
+debug = True
+diff_colormap = "BrBG"
+output_format_subplot = ["png"]
+regions = ["20S20N"]
+results_dir = "all_sets_results_test"
+seasons = ["ANN"]
+test_start_yr = '2012'
```
forsyth2: @chengzhuzhang Added a test that has all of `{test, ref}_{start, end}_yr` set.

tests/system/all_sets.cfg
```diff
@@ -168,11 +168,51 @@ print_statements  = True
...
-plot_type = "scatter"
-case_id = "TREFHT"
+case_id = 'TREFHT-response-map-start-yrs'
 debug = True
+diff_colormap = "BrBG"
+output_format_subplot = ["png"]
+regions = ["20S20N"]
 results_dir = "all_sets_results_test"
+seasons = ["ANN"]
+test_start_yr = '2012'
```
forsyth2: @chengzhuzhang Keeping this test, as it checks that we can mix and match the different `{start, end}_yr` parameters.

acme_diags/derivations/acme.py
chengzhuzhang: Good idea.

chengzhuzhang:
``` 
 'NET_FLUX_SRF': OrderedDict([
        (('FSNS','FLNS','LHFLX','SHFLX'), lambda fsns, flns, lhflx, shflx: netflux4(fsns, flns, lhflx, shflx)),
        (('rsds','rsus','rlds','rlus','hfls','hfss'), lambda rsds, rsus, rlds, rlus, hfls, hfss: netflux6(rsds, rsus, rlds, rlus, hfls, hfss)),
```
I meant to add an entry
(('FSNS','FLNS',**'QFLX',**'SHFLX'), lambda fsns, flns, lhflx, shflx: netflux4(fsns, flns, lhflx, shflx)),

with this netflux4 function needs to take in account the qflx to lflx conversion. Or we can have a new netflux4 function just for take care of this.

acme_diags/derivations/acme.py
```diff
@@ -97,6 +97,14 @@ def qflxconvert_units(var):
         var.units = 'mm/day'
     return var
 
+
+def qflx_convert_to_lhflx(qflx):
```
forsyth2: @chengzhuzhang I use this function to help derive `NET_FLUX_SRF` on Line 533 and `LHFLX` on Line 622. If the changes in this file (acme.py) look good to you, then this pull request can be merged.

acme_diags/driver/enso_diags_driver.py
```diff
@@ -59,10 +67,10 @@ def calculate_nino_index_model(data, nino_region_str, parameter):
             sst = data.get_timeseries_variable('SST')
         except RuntimeError as e1:
             if str(e1).startswith('Neither does SST nor the variables in'):
-                print('Handling exception=', e1)
+                print('Handling the following exception by looking for surface temperature:', e1)
```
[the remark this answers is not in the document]
forsyth2: Agreed that `TS` is the surface temp. Resolving this conversation.

PR #301, opened 2020-05-05
Wind vectors

forsyth2:
Add wind vectors diagnostic to e3sm_diags.


forsyth2:
Note to self: do NOT squash later commits into the first one until this pull request is ready to merge. That way, it will be easy to tell what changes I've made to the original version (i.e., the first commit).

chengzhuzhang:
@forsyth2 Thank you for working on reviving this branch!
Under `test_model_data_for_acme_diags/climatology`, you should be able to find netcdf data having names with the case name (20160520.A_WCYCL1850.ne30_oEC.edison.alpha6) as the first part of the filename. `U,V,TAUX,TAUY`should be all included within those files.

I don't remember I sew those Errors you mentioned here.

For the questions:
1. The super long arrows, it's not an error, but we do need an algorithm to scale them so that they can be presented nicely. The algorithm should be flexible for data at different resolutions (i.e, high resolution and low resolution). This is where my progress stopped during my attempt :(

2. 2a, Climate models often report the u and v components (zonal and meridional wind, respectively) of the wind vector. U and V are output at each grid point. But for visualizing the vectors, it would be too dense to show one vector per grid, therefore regridding is desired. 
2b, TAUX and TAUY is different that the TAU we used for cloud optical depth. The Greek letter tau is used quite often as in our case here, its just a coincident.
2c, Surface wind stress is a dragging force applied to the wind blow at surface. it has a unit N/m2. A detailed explanation can be found here: https://marine.rutgers.edu/dmcs/ms501/2004/Notes/Wilkin20041014.htm

3. we don't need more to add for this set. But same plot with a polar projection zooming in over N. and S. Polar areas will be desired. That would be another PR I suppose.

acme_diags/driver/lat_lon_vector_driver.py
```diff
@@ -0,0 +1,248 @@
...
+    
+                        # Regrid towards the lower resolution of the two
+                        # variables for calculating the difference.
+                        mv1_reg, mv2_reg = utils.general.regrid_to_lower_res(
+                            mv1_domain, mv2_domain, parameter.regrid_tool, parameter.regrid_method)
+
+                        tests.append(mv1_reg)
+                        refs.append(mv2_reg)
+
+                        # TODO: Do we want to save these json files? Net CDF files?
```
forsyth2, line 126: Should we save any json or Net CDF files?

acme_diags/driver/utils/dataset.py
```diff
@@ -278,6 +278,10 @@ def _find_climo_file(self, path_name, data_name, season):
         """
         Locate climatology file name based on data_name and season.
         """
+        if os.path.isfile(path_name):
+            # Without this line, lat_lon_vector plots fail to show seasonal plots.
+            # TODO: However, the seasonal plots look identical to the ANN plots. Why?
```
forsyth2, line 283: The seasonal plots look identical to the annual plots. Is something set up wrong?

acme_diags/plot/cartopy/lat_lon_vector_plot.py
```diff
@@ -0,0 +1,234 @@
...
+                     cmap=cmap,
+                     extend='both',
+                     )
+    
+    ax.set_aspect('auto')
+    ax.coastlines(lw=0.3)
+
+    if n != 2:
+        # Normalized vector fields for test and ref, but not diff
+        # TODO: Scale the vectors properly.
```
forsyth2, line 85: I tried changing the stride (e.g., the `5` in `x[::5, ::5]`), but that didn't look good either. I'm unclear on how to adjust the stride algorithmically based on different lat/lon. The plots linked on this review do not scale the vectors in any way.

acme_diags/viewer/default_viewer.py
```diff
@@ -73,22 +80,25 @@ def create_viewer(root_dir, parameters):
...
-                            row_name = '{} {}'.format(var, region)
+                            row_name = '{} {}'.format(var_title, region)
                         else: 
-                            row_name = '{} {} {}'.format(var, region, ref_name)
+                            row_name = '{} {} {}'.format(var_title, region, ref_name)
                         fnm = '{}-{}-{}-{}'.format(ref_name,
-                                                    var, season, region)
+                                                    var_title, season, region)
+                        print('DEBUG: Row name A: {}'.format(row_name))
+                        # TODO: Come up with better row names
```
forsyth2, line 89: What would be a better row name?
For reference, in the tests, the row name for `lat_lon_vectors` is `U-V-850mb global ERA-Interim` and the row name for `lat_lon` is `T-850mb global ERA-Interim`. Those row names seem more or less equivalent in what information they give.

run.py
```diff
@@ -0,0 +1,52 @@
+# TODO: Remove before merging
```
forsyth2, line 1: Remove this script before merging

tests/system/all_sets.cfg
```diff
@@ -85,6 +85,25 @@ backend = "mpl"
...
+case_id = "ERA-Interim"
+variables = ["U", "V"]
+seasons = ["ANN"]
+plevs = [850.0]
+contour_levels = [240, 245, 250, 255, 260, 265, 270, 275, 280, 285, 290, 295]
+diff_levels = [-10, -7.5, -5, -4, -3, -2, -1, -0.5, 0.5, 1, 2, 3, 4, 5, 7.5, 10]
+backend = "mpl"
+results_dir = "all_sets_results_test"
+debug = True
+# TODO: Replace with paths to external testing directory accessible by CircleCI.
```
forsyth2, line 99: What directory should I use for the `data_path` variables? It would need to be something easily accessible or downloadable (e.g., wouldn't ask for a password on every use).

acme_diags/driver/lat_lon_vector_driver.py
chengzhuzhang: Yes, if save_netcdf is on, we should save at lease wind speed as netcdf. And the metrics based on wind speed should be saved following other lat-lon variables.

acme_diags/viewer/default_viewer.py
chengzhuzhang: `U-V-850mb global ERA-Interim` is a good name

tests/system/all_sets.cfg
chengzhuzhang: Can we follow what CDAT does? https://github.com/CDAT/cdat_info/blob/4d35272dab39d162ae1cd9d99a6aa03e5dbdb1b9/cdat_info/cdat_info_src.py#L108

We can place our testing data on our data server at anl. 
https://web.lcrc.anl.gov/public/e3sm/
More info: https://acme-climate.atlassian.net/wiki/spaces/ED/pages/707002387/E3SM+Input+Data+Servers

@zshaheen I vaguely remember you started similar effort making CI work. Could you chime in on this? Thanks

acme_diags/driver/utils/dataset.py
chengzhuzhang: I think the following lines in `run.py` caused the problem:

```
param.test_data_name = '20161118.beta0.F1850COSP.ne30_ne30.edison_ANN_climo.nc'
param.ref_data_name = '20161118.beta0.F1850COSP.ne30_ne30.edison_ANN_climo.nc'
param.test_name = '20161118.beta0.F1850COSP.ne30_ne30.edison_ANN_climo.nc'
param.ref_name = '20161118.beta0.F1850COSP.ne30_ne30.edison_ANN_climo.nc' 
```
This way the input data were hard-coded to use ANN for all. Replace those with below two lines, should work...
```
param.test_name = '20161118.beta0.F1850COSP.ne30_ne30.edison`
param.ref_name = '20161118.beta0.F1850COSP.ne30_ne30.edison`

File: tests/system/all_sets.cfg (hunk 6 above)
Comment (zshaheen) [reply tests/system/all_sets.cfg:99]: Hmm, I don't recall anything like what you're asking about.

Like what you said, I remember trying to setup the test data on another DOE machine and have all of the downloading be from there. I tired to get something similar to how MPAS-Analysis did it. The data is broken into chunks and transferred over piecewise, allowing for a cool download bar denoting the progress.

File: acme_diags/driver/lat_lon_vector_driver.py [hunk 7]
```diff
@@ -0,0 +1,197 @@
...
+        tests.append(test)
+        refs.append(ref)
+        diffs.append(tests[ZONAL] - ref[ZONAL])
+        diffs.append(tests[MERIDIONAL] - ref[MERIDIONAL])
+        # Note that diff is not the vector norm of the above two components.
+        # It is the difference of the test and ref vector norms.
+        diffs.append(diff)
+        #result = Results_Container(tests=tests, refs=refs, diff=diff, metrics = metrics_dict)
+
+        # Saving the metrics as a json.
```
forsyth2, line 168: This code block appears to be sufficient to save the metrics dict to `json`. I removed the commented-out code blocks in the `if/else` for z-axis above.

tests/system/all_sets.cfg
forsyth2: See lines 60-61 in `tests/system/test_diags.py`. I will update the data path to be Anvil once I get access to that machine. Alternatively, we could host the test data on Cori. 

Also, it might be a good idea to have another pull request to move all the other `nc` files out of this repo and onto Anvil/Cori, since that would be outside the scope of this PR. I know #303 requires removing the `nc` files from the entire git history (i.e., not just the current state), but I believe it would be good to make sure the repo has no `nc` files in its current state before attempting that.

acme_diags/plot/cartopy/lat_lon_vector_plot.py
forsyth2: @chengzhuzhang if the pull request looks good (aside from the test data path), I can hand it back over to you to continue trying to scale the vectors properly.

chengzhuzhang:
@forsyth2 The PR looks good. I will take it from here and try finalizing the vectors. Thank you.

tests/system/all_sets.cfg
forsyth2: @chengzhuzhang Yeah the paths are set in the `cfg` files, so I just added a test data path and set the test to copy over the data to that path. Right now, it's set to copy from another of my directories outside the repo. So, the "copy over" step (lines 60-61 in `tests/system/test_diags.py`) needs to be changed to copy down from Anvil instead.

zshaheen: Yeah guys sorry, I don't really recall much regarding this. It's been some time.

chengzhuzhang:
@wlin7 As we discussed yesterday, we should make a push to revive and complete this PR. The current block is that the visualization of arrows for wind vectors are not optimized. If you could review the current example and make change suggestions that would be great. https://portal.nersc.gov/project/e3sm/forsyth/wind_vectors_2020_06_15/wind_vectors_results/viewer/

forsyth2:
Consider streamlines (https://matplotlib.org/stable/gallery/images_contours_and_fields/plot_streamplot.html) rather than wind vectors

wlin7:
Sounds great, @forsyth2 . It is a better and flexible alternative in many ways. Let's try a few options with our typical global array. The views may look different than the examples as our global grids are much denser.

PR #308, opened 2020-07-01
Add two ERA5 post processing scripts

chengzhuzhang:
This PR should replace PR#293. 
- One script specifically for ERA5 U wind component, which has denser plevs above 100mb available than current ERA-interim data set, in preparation for QBO diags.(1 deg)
-One script for standard ERA5 variables from CREATE-IP project. (0.25 deg)


chengzhuzhang:
These pre-processing scripts are mostly independent from e3sm_diags main code base.  I will merge without code review.

PR #309, opened 2020-07-08
Set QBO output file automatically

forsyth2:
Other diagnostics, such as ENSO diags, automatically set the `output_file` so that the user doesn't have to do so. This PR adds this functionality to QBO diagnostics.

### DIFF
--- a/acme_diags/driver/qbo_driver.py
+++ b/acme_diags/driver/qbo_driver.py
@@ -194,6 +194,7 @@ def run_diag(parameter):
                     test_json[key] = list(test[key])
                     ref_json[key] = list(ref[key])
 
+        parameter.output_file = 'qbo_diags'
         # TODO: Check the below works properly by using ncdump on Cori
         utils.general.save_transient_variables_to_netcdf(parameter.current_set, test_nc, 'test', parameter)
         utils.general.save_transient_variables_to_netcdf(parameter.current_set, ref_nc, 'ref', parameter)


acme_diags/driver/qbo_driver.py
```diff
@@ -194,6 +194,7 @@ def run_diag(parameter):
                     test_json[key] = list(test[key])
                     ref_json[key] = list(ref[key])
 
+        parameter.output_file = 'qbo_diags'
```
forsyth2, line 197: Other diags have variables put into the output file name, but QBO diags seems fairly independent of variables (e.g., variable is always `U`,`V`, etc.) . Are there any we should include? Compare to ENSO diags: https://github.com/E3SM-Project/e3sm_diags/blob/master/acme_diags/driver/enso_diags_driver.py#L270

chengzhuzhang: U is the only variable we will apply for QBO. In my test, the variable name is not shown up in the viewer though: https://acme-viewer.llnl.gov/zhang40/qbo_diags_acme1_m2m/viewer/qbo/index.html.

chengzhuzhang: I haven't tested with the new change though.

forsyth2: @chengzhuzhang Oh that's interesting. I see `ERA-Interim` under `Variable` when I use my `run` script. 

Should I add `U` to the `output_file` name?

chengzhuzhang: It seems like case-id is shown in your viewer instead of variable? I was running a model vs model case.

forsyth2: Well, ENSO uses `case_id` too: https://github.com/E3SM-Project/e3sm_diags/blob/master/acme_diags/driver/default_diags/enso_diags_model_vs_model.cfg#L3, https://github.com/E3SM-Project/e3sm_diags/blob/master/acme_diags/viewer/enso_diags_viewer.py#L59

chengzhuzhang: I found another problem with plotting. And will try tweak the viewer option some. I will update in another PR.

forsyth2:
Included in #310

PR #310, opened 2020-07-10
Qbo fix

chengzhuzhang:
A couple fixes for QBO Diags:
-Add function to unify plevs.
-Inverse plev axes.
-Updates default cfg files. 
-Bug fix 

Example output:
Model vs obs:https://acme-viewer.llnl.gov/zhang40/qbo_diags_acme1/viewer/
Model vs model: https://acme-viewer.llnl.gov/zhang40/qbo_diags_acme1_m2m/viewer/

### DIFF
--- a/acme_diags/driver/qbo_driver.py
+++ b/acme_diags/driver/qbo_driver.py
@@ -12,17 +12,17 @@ import MV2
 
 def unify_plev(var):
     """
-    Given a data set with a z-axis, 
-    convert to the plev with units: hPa and make sure z is positive up (same as model data)
+    Given a data set with a z-axis (plev), 
+    convert to the plev with units: hPa and make sure plev is in ascending order(same as model data)
     """
     var_plv = var.getLevel()
     if var_plv.units == 'Pa':
         var_plv[:] = var_plv[:]/100.0 #convert Pa to mb
         var_plv.units = 'hPa'
         var.setAxis(1, var_plv)
-    # For positive down plevel: from TOP (level 0) to BOTTOM (last level), i.e Plev value
-    # going up with each level", revers it. 
-    if var.getLevel()[0] < var.getLevel()[-1]:
+
+    #Make plev in ascending order
+    if var.getLevel()[0] > var.getLevel()[-1]:
         var = var(lev=slice(-1, None, -1))
 
 
--- a/acme_diags/driver/qbo_driver.py
+++ b/acme_diags/driver/qbo_driver.py
@@ -89,7 +89,6 @@ def process_u_for_power_spectral_density(data_region):
     level_top = 18
     # Average over lat and lon
     data_lat_lon_average = cdutil.averager(data_region, axis='xy')
-    level_data = data_lat_lon_average.getAxis(1)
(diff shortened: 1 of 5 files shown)


chengzhuzhang:
@forsyth2 This PR included the change in https://github.com/E3SM-Project/e3sm_diags/pull/309 Please let me know if you catch any problems. I will make a new release after merging this PR. Thank you.

acme_diags/driver/qbo_driver.py
```diff
@@ -8,6 +8,22 @@
...
+    Given a data set with a z-axis, 
+    convert to the plev with units: hPa and make sure z is positive up (same as model data)
+    """
+    var_plv = var.getLevel()
+    if var_plv.units == 'Pa':
+        var_plv[:] = var_plv[:]/100.0 #convert Pa to mb
+        var_plv.units = 'hPa'
+        var.setAxis(1, var_plv)
+    # For positive down plevel: from TOP (level 0) to BOTTOM (last level), i.e Plev value
+    # going up with each level", revers it.
```
forsyth2, line 24: I think a better wording might be `if z is positive down (level 0 < last level), then reverse z`.

acme_diags/plot/cartopy/qbo_plot.py
```diff
@@ -69,19 +71,23 @@ def plot(period_new, parameter, test, ref):
     z = dict(data=test['qbo'].T)
     ax0 = plot_panel(0, fig, 'contourf', label_size, '{} U 5S-5N'.format(test['name']), x, y, z=z,
                plot_colors=cmap2, color_levels=color_levels0, color_ticks=[-50, -25, -5, 5, 25, 50])
+    ax0.invert_yaxis()
```
forsyth2, line 74: I'm not quite sure why we're inverting these axes. Looking at the graphs, pressure seems to be increasing from bottom to top (which would match with your `unify_plev` function); however, since pressure decreases with altitude, I thought the convention was to have pressure decrease going up the y axis? Or am I misinterpreting?

chengzhuzhang: The convention showing the vertical profiles is to have pressure decreases going up the y axis (i.e. altitude increases with y). I have pressure decreasing from bottom to top https://acme-viewer.llnl.gov/zhang40/qbo_diags_acme1/viewer/qbo/bvariable/bqbo-era5/bplot.html
Would you post your testing results?

acme_diags/driver/qbo_driver.py
chengzhuzhang: I changed the wording to be more clear..

acme_diags/plot/cartopy/qbo_plot.py
chengzhuzhang: Thank you for catching this! It took some time to troubleshooting. It turns out that, the 3.1.0 version of matplotlib was causing the trouble. After I upgrade to 3.1.2. I don't need to use `ax.invert_yaxis()` to see the correct pressure levels. I deleted these three lines in the new commit.

PR #313, opened 2020-07-14
fix for conda build

chengzhuzhang:


xylar:
@chengzhuzhang, I'm testing with my suggested changes right now.  I'll let you know if it looks like I expect when I do `conda search --info --use-local e3sm_diags=2.1.0`.

conda/meta.yaml
```diff
@@ -1,16 +1,18 @@
...
-    version: 2.1.0
+    name: {{ name|lower }}
+    version: {{ version }}
 
 source:
-    #path: ..
-    git_url: https://github.com/E3SM-Project/e3sm_diags
-    git_rev: v2.1.0
+    url: https://github.com/E3SM-Project/e3sm_diags/archive/v{{ version }}.tar.gz
+    sha256: 9c169c27fcdfec870b4e20fd71c24a3a426136424c32003bb9a714af157a3cb5
```
xylar, line 10: Since the tarball has been replaced, you will now need to update the sha256.  That's just how it goes and one reason conda-forge keeps the recipes separate from the code itself.

xylar: Oh, sorry, I though your release corresponded to updating the v2.1.0 tag but it looks like you left it the same, which I think is good.

You patch the code in the meta.yaml to include any of the fixes you've made since then if you need to.  Let me know if you want help with that.

PR #317, opened 2020-07-20
Patch the conda build with the new version number

xylar:
Commit 8b3b50b804aba393896addcd90fa740ae338d70e went in after the v2.1.0 release but is needed so E3SM_Diags includes the correct version number in the metadata.


xylar:
@chengzhuzhang, I just build this and uploaded it to the `e3sm` channel so we should be set for this to get merged when you're ready.

PR #323, opened 2020-08-06
update conda env ymls for issue 319

### LINKED ISSUE #319: Not able to create a dev environment following docs
I'm not able to able to follow the instructions for creating an environment using the file https://github.com/E3SM-Project/e3sm_diags/blob/master/conda/e3sm_diags_env_dev.yml
`conda create` produces the usual unhelpful message about conflicts:
```
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
- failed
```
It seems maybe the `yaml` file needs to be updated.

chengzhuzhang:
The yml files are created on a Linux machine based on:
```
conda create -n e3sm_diags_env_2.1.1 e3sm_diags=2.1.1 python=3 mesalib -c e3sm -c conda-forge -c cdat/label/v82
conda env export > e3sm_diags_env.yml --no-builds
```

closes #319


xylar:
Following the (strong) recommendation of [conda-forge](https://conda-forge.org/#about), I have channel order set to strict:
```
conda config --set channel_priority strict
```
This ensures that packages come from higher priority channels if available and that lower priority channels don't get searched.

As a result, of setting channel priority to strict, the order of the channels in the yaml files is important.  It should be:
```
  - conda-forge
  - defaults
  - e3sm
  - cdat/label/v82
```
(I don't think the order of the last two matters because there should be no packages in common between e3sm and cdat.)

Only with this change am I able to create these environments.

xylar:
I'm still having trouble with the `dev` version:
```
['/home/xylar/miniconda3/envs/test/bin/python', '-m', 'pip', 'install', '-U', '-r', '/home/xylar/Desktop/condaenv.29b8ry7x.requirements.txt']
Pip subprocess output:

Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement cdms2==3.1.4 (from -r /home/xylar/Desktop/condaenv.29b8ry7x.requirements.txt (line 1)) (from versions: none)
ERROR: No matching distribution found for cdms2==3.1.4 (from -r /home/xylar/Desktop/condaenv.29b8ry7x.requirements.txt (line 1))


CondaEnvException: Pip failed
```

xylar:
I think those `pip` packages are not correct.  Something is going wrong on `conda-forge` and those are getting labeled as coming from `pip` when they are actually from `conda-forge`.  I think it would be better to move them out of the `pip` section.

xylar:
It might be worth posting an issue on the feedstocks for those two as well.  I'll do that.

xylar:
Okay, one more thing: It seems that `regrid2` is not an independent package and gets installed along with `cdms2`.  This is not good (https://github.com/conda-forge/cdms2-feedstock/issues/61) but probably hasn't caused trouble until this yaml generation issue makes it clear.  The result is that `regrdi2` needs to be taken out of the list of packages because `conda` will not be able to find it.

xylar:
Nope, even with all of that, I'm not able to get it to work for me.  I will try to create my own environment with your commands:
```
conda create -n e3sm_diags_env_2.1.1 e3sm_diags=2.1.1 python=3 mesalib -c e3sm -c conda-forge -c cdat/label/v82
conda env export > e3sm_diags_env.yml --no-builds
```
and see what is different.

xylar:
Okay, I think I have it figured out.  My guess is that without strict channels, you are picking up `cdms2` (and maybe other packages) from `cdat/label/v82` instead of `conda-forge`.  We always want the `conda-forge` versions for any packages on conda-forge.

I was able to get things working by:
* changing the channel order as in my comment above
* moving `cdms2` out of `- pip` and into the main list
* removing `regrid2`
* either:
  * switching to `cdms2=3.1.5` or
  * switching to `libopenblas=0.3.6` and `openblas=0.3.6`

I think there is not a bild of `cdms2=3.1.4` on conda-forge that works with the newer blas (0.3.10) that you are pinning, whereas there must be one on `cdat/label/v82`.

chengzhuzhang:
Thank you for testing and the suggestions! I'm a little nervous to `remove regrid2` as a dependency, because that was used as a regrid option.

xylar:
I'm pretty darn sure that regrid2 is part of the cdms2 install, even though it shouldn't be. You can take it out and verify for yourself that it gets installed anyway. But you can also leave it because pip will just see that it's already installed. But there is no pypi package for it so it can't actually be installed with pip.

chengzhuzhang:
Updated the ymls according to @xylar's comments, and tested the envs created based these two ymls. All tests passed.

chengzhuzhang:
@xylar Thank you for testing!

PR #332, opened 2020-08-17
Update docs gitignore

forsyth2:
Update `.gitignore` on `gh-pages` branch. After this pull request is merged, `master` and `gh-pages` will have identical `.gitignore` files. Previously files that were ignored on `master` (and branches from it) would show up as changed when switching to `gh-pages` (and branches from it). This pull request should stop this.


forsyth2:
@chengzhuzhang I noticed on #331 that many files that should have been ignored were included. Turns out the `.gitignore` on `gh-pages` was much smaller than the one on `master`.

Is there any reason we should ignore/no ignore different file groups on `master` and `gh-pages`? If not, I will merge this.

chengzhuzhang:
Ryan, please go ahead and merge. Thanks for noticing it.

xylar:
@chengzhuzhang and @forsyth2, typically, your `gh-pages` branch wouldn't include the code itself, so the `.gitignore` wouldn't need to be as complicated.  Here are some examples from our repos:
https://github.com/MPAS-Dev/MPAS-Analysis/tree/gh-pages
https://github.com/MPAS-Dev/MPAS-Tools/tree/gh-pages
https://github.com/MPAS-Dev/MPAS-Model/tree/gh-pages

Is there a reason you include the code on your `gh-pages` branch, too?

chengzhuzhang:
@xylar I agree. It would be cleaner not to include codes. @forsyth2 maybe instead of having a long `.gitignore`, we should follow the examples Xylar provided?

forsyth2:
@chengzhuzhang The second commit removes unnecessary files. I will combine the commits before merging.

`ls` now gives:
```
Makefile            e3sm_diags.egg-info source
docs                index.html
```
I'm not sure if `e3sm_diags.egg-info` needs to be there. The other files/directories I think all need to be present. 

I'm not entirely sure which lines I can remove from the `.gitignore` though. For example, the new `.gitignore` includes `build/` and `dist/`, which I removed in the second commit. I created another branch `remove-unncessary-files` with just the second commit (i.e., no `.gitignore` changes), ran the tests on `master` and changed to the `remove-unncessary-files` branch. That caused `tests/` to show up in `git status`. Turns out that directory has `__pycache__` in it, which implies the `.gitignore` needs to _at least_ have `__pycache__/` in it.

chengzhuzhang:
Hi Ryan,  I can't seem to find the `remove-unncessary-files` branch. Would you push the branch? With `__pycache__/` included in .gitignore will work fine I think.

forsyth2:
@chengzhuzhang I didn't push the `remove-unnecessary-files` branch to GitHub -- it just had the second commit of this PR (it did not include _any_ changes to the `.gitignore`). I suppose I can update the `.gitignore` only with `__pycache__`. If it turns out later we should have included more in the `.gitignore`, then we can just add them then.

forsyth2:
@chengzhuzhang Ok, I updated the `.gitignore` just to add `__pycache__/`. Everything else in the PR is deleting files we don't need. I think this should be good to merge then.

forsyth2:
This should be good to merge. We can add more files to the `.gitignore` later, if needed. This PR deletes temporary files (ending in `~`), `acme_diags`, `build`, `dist`, `e3sm_diags_env_dev.yml`,  and `tests`. It also includes changes resulting from running `make html`.

PR #355, opened 2020-11-11
Improve Streamflow Performance

forsyth2:
Improve Streamflow Diags performance when using `.nc` files.

### DIFF
--- a/acme_diags/driver/streamflow_driver.py
+++ b/acme_diags/driver/streamflow_driver.py
@@ -3,8 +3,10 @@ from __future__ import print_function
 import csv
 import numpy
 import scipy.io
-import time
+
+import cdms2
 import cdutil
+
 from acme_diags.driver import utils
 from acme_diags.plot.cartopy.streamflow_plot import plot_seasonality_map, plot_annual_map, plot_annual_scatter
 
@@ -34,6 +36,7 @@ def get_drainage_area_error(radius, resolution, lon_ref, lat_ref, area_upstream,
 
 
 def get_seasonality(monthly):
+    monthly = monthly.astype(numpy.float64)
     # See https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2018MS001603 Equations 1 and 2
     if monthly.shape[0] != 12:
         raise Exception('monthly.shape={} does not include 12 months'.format(monthly.shape))
@@ -48,7 +51,8 @@ def get_seasonality(monthly):
         streamflow_month_all_years = monthly[month, :]
         # Proportion that this month contributes to streamflow that year.
         # 1 x num_years
-        streamflow_proportion = streamflow_month_all_years / total_streamflow
+        # For all i, divide streamflow_month_all_years[i] by total_streamflow[i]
+        streamflow_proportion = numpy.divide(streamflow_month_all_years, total_streamflow)
         # The sum is the sum over j in Equation 1.
         # Dividing the sum of proportions by num_years gives the *average* proportion of annual streamflow during
         # this month.
@@ -158,6 +162,8 @@ def run_diag(parameter):
         if parameter.print_statements:
             # For edison: 720x360x600
             print('test_array.shape={}'.format(test_array.shape))
+        if type(area_upstream) == cdms2.tvariable.TransientVariable:
+            area_upstream = area_upstream.getValue()
 
         # Resolution of MOSART output
         resolution = 0.5
@@ -174,12 +180,6 @@ def run_diag(parameter):
         export = numpy.zeros((lat_lon.shape[0], 9))
         if parameter.print_statements:
             print('export.shape={}'.format(export.shape))
-        drainage_area_error_time = 0
-        sum_time = 0
-        mean_time = 0
-        monthly_mean_time = 0
-        seasonality_time = 0
-        t_initial = time.time()
         for i in range(lat_lon.shape[0]):
             if parameter.print_statements and (i % 1000 == 0):
                 print('On gauge #{}'.format(i))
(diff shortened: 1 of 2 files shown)


demo_run.py
```diff
@@ -0,0 +1,47 @@
+# TODO: Remove before merging
```
forsyth2, line 1: Remove before merging

tests/system/all_sets.cfg
```diff
@@ -283,6 +283,9 @@ max_num_gauges = 12
 #gauges_path =
 #test_data_path =
 #ref_mat_file =
+gauges_path = '/Users/forsyth2/projectAIMS/E3SM-Project/diags_collected/For_Jill/GSIM_catchment_characteristics_all_1km2.csv'
+test_data_path = '/Users/forsyth2/projectAIMS/data/streamflow_data3/'
+ref_mat_file = '/Users/forsyth2/projectAIMS/E3SM-Project/diags_collected/For_Jill/GSIM.mat'
```
forsyth2, line 288: Remove before merging.

tests/test_run.py
```diff
@@ -51,6 +51,8 @@ def test_all_sets_and_all_seasons(self):
         # TODO: Replace these paths with paths that automated testing tools can access (in future pull request).
         #streamflow_param.gauges_path =
         #streamflow_param.ref_mat_file =
+        streamflow_param.gauges_path = '/Users/forsyth2/projectAIMS/E3SM-Project/diags_collected/For_Jill/GSIM_catchment_characteristics_all_1km2.csv'
+        streamflow_param.ref_mat_file = '/Users/forsyth2/projectAIMS/E3SM-Project/diags_collected/For_Jill/GSIM.mat'
```
forsyth2, line 55: Remove before merging.

forsyth2:
I visually checked the plots produced from the `nc` file before and after these changes. They appear identical.

Runtimes (in seconds, rounded to one decimal place) after these changes:
|Step|`nc` file|`mat` file|diff|
|---|---|---|---|
|`time python demo_run.py` real time| 77.5|50.3|27.2|
|Time to loop through 25,000+ gauges|57.6|26.8|30.8|
|`get_drainage_area_error`|1.6|1.6|0.0|
|`monthly_mean`|13.2|13.2|0.0|
|`monthly_mean`: `sum`|1.9|1.8|0.1|
|`monthly_mean`: `mean`|10.2|10.3|-0.1|
|`seasonality`|8.5|8.5|0.0|
|`seasonality`: Loop line 1|0.2|0.2|0.0|
|`seasonality`: Loop line 2|1.0|0.9|0.1|
|`seasonality`: Loop line 3|5.7|5.7|0.0|

forsyth2:
@chengzhuzhang The relevant code changes are all in `acme_diags/driver/streamflow_driver.py`. I posted runtimes in the above comment. How do those times look? Thanks!

chengzhuzhang (commented):
Thank you for making the performance improvement. The code change look good to me.
I guess it will be cleaner to remove the lines for timing info. Also I'm curious what's the real time to do a say 20 years run?

chengzhuzhang:
I used to use a python tool called 'cProfile' for measuring performance. Maybe it will be handy to try to use it?

forsyth2:
@chengzhuzhang The timing info only prints if the user specifies `streamflow_param.print_statements = True`, so I don't really have a problem keeping those lines in for debugging. That said, I'm not really sure if those lines slow down the code any (the time code still runs even if that parameter is turned off).

Is there a set of 20 years of `RIVER_DISCHARGE_OVER_LAND_LIQ` I can try running on?

zshaheen:
Just my $0.02, but I would strongly suggest you use `cProfile`, as mentioned by @chengzhuzhang.

With my work for `e3sm_diags`, I used  `cProfile` to see what functions took long and [`line_profiler`](https://github.com/pyutils/line_profiler) to see what line took the longest. These are a more complete solution when analyzing the performance of your code. It'll save time rather than manually inserting calls from `time`. It also is a cleaner solution in my opinion, since most users of `e3sm_diags` won't really care about the timing info.

Do note that when profiling your code with the aforementioned tools, it took about 10-100x longer to run if I recall correctly.

chengzhuzhang:
Thanks for chiming in @zshaheen ! Yes, I do recall `line_profier` has been helpful. Utilizing these tools both saves time and makes the codes cleaner.

 @forsyth2 I thought you had a 500 years data file earlier `RIVER_DISCHARGE_OVER_LAND_LIQ_000101_050012.nc`?

acme_diags/driver/streamflow_driver.py
```diff
@@ -303,9 +333,15 @@ def run_diag(parameter):
                     monthly = numpy.ones((12, 1))
                 else:
                     monthly = mmat
+
+                if type(monthly) == cdms2.tvariable.TransientVariable:
```
forsyth2, line 337: (From https://github.com/pyutils/line_profiler) Using `kernprof -l demo_run.py` and `python -m line_profiler demo_run.py.lprof`, I found that the lines above take a while to run:
```
ref = numpy.squeeze(ref_array[ref_lon-1, ref_lat-1, :])
mmat = numpy.transpose(numpy.reshape(ref, (-1,12)))
monthly_mean_ref = numpy.nanmean(mmat, axis=1)
annual_mean_ref = numpy.mean(monthly_mean_ref)
```
They take 23.2, 3.2, 27.6, and 8.2 seconds respectively. Unfortunately, I couldn't move this `TransientVariable` check any further up because it would cause the Scatter Plot to have a `R^2` of 0 or take longer.

forsyth2:
@chengzhuzhang Thanks, I forgot about that file. I ran with 20 years and with all 500 years. The time to process all 25,000+ gauges stays pretty consistently around 60 seconds independent of number of years, which confirms my thoughts from a few weeks ago.

Total processing time for 500 years was under five minutes (it does take longer to read 500 years into memory than 20).

chengzhuzhang:
@forsyth2 Thanks for confirming. Five mins processing time for 500 years is very good.  I think this PR is ready to merge after some cleanup.

PR #357, opened 2020-11-21
Streamflow Diags Enhancements

forsyth2:
Streamflow Diags enhancements

- Specifications for data paths for test and ref `.nc` and `.mat` files.
- Improve plot titles.

PR #359, opened 2020-11-25
Fix QBO Names

forsyth2:
Fix how QBO code handles test/ref names.


forsyth2:
@chengzhuzhang This pull request (along with #357 for Streamflow Diags) will make it so that all the diags use some variant of `get_name_and_yrs`. QBO doesn't use that particular function because the plot titles also include `U 5S-5N`.

demo_run.py
```diff
@@ -0,0 +1,44 @@
+# TODO: Remove before merging
```
forsyth2, line 1: Remove before merging.

chengzhuzhang (approved):
I haven't actually tested this. But the codes look fine.

PR #370, opened 2020-12-16
Image Comparison Tests

### LINKED ISSUE #328: Create automated tests for completeness of figures and comparing figures.

forsyth2:
Add image comparison tests. Resolves #328.


tests/system/test_diags.py
```diff
@@ -65,26 +75,42 @@ def setUpClass(self):
...
+            path_to_actual_png = full_png_path
+            path_to_expected_png = '../unit_test_images/{}'.format(png_path)
+            #print('path_to_actual_png={}'.format(path_to_actual_png))
+            #print('path_to_expected_png={}'.format(path_to_expected_png))
+            actual_png = Image.open(path_to_actual_png).convert('RGB')
+            expected_png = Image.open(path_to_expected_png).convert('RGB')
+            diff = ImageChops.difference(actual_png, expected_png)
+
+            # If `diff.getbbox` is None, then the images are in theory equal
+            self.assertIsNone(diff.getbbox())
```
forsyth2, line 111: Compares two images to determine if they're identical.

tests/system/test_diags.py
```diff
@@ -186,12 +213,18 @@ def check_streamflow_plots(self):
...
 
     # Test results_dir
     def test_results_dir(self):
         self.assertTrue(TestAllSets.results_dir.endswith('all_sets_results_test/'))
 
+    # Test the image count
+    def test_image_count(self):
+        actual_num_images = count_images('all_sets_results_test')
+        expected_num_images = count_images('../unit_test_images')
+        self.assertEqual(actual_num_images, expected_num_images)
```
forsyth2, line 226: Check that we have the correct number of images generated.

tests/system/test_diags.py
```diff
@@ -239,8 +272,12 @@ def test_enso_diags_scatter_test_ref_yrs(self):
         case_id = 'TREFHT-response-scatter-test-ref-yrs'
         self.check_enso_scatter_plots(case_id)
 
+    # Image diff for `test_lat_lon` fails sporadically, so setting `image_diff=False`.
```
forsyth2, line 275: For some reason, `test_lat_lon` fails sporadically on the `image_diff` check. I'm not sure why. `diff.show()` doesn't show any differences between `actual_png` and `expected_png`. Sometimes, the check does in fact pass. So, for now, I'm just disabling that check and we can visually inspect that one plot.

forsyth2 (commented):
@chengzhuzhang This pull request is ready for review. See comments below for more details. Thanks!

tests/system/test_diags.py
chengzhuzhang: I suppose actual_png and expected_png are identical in this test? It's kind of odd the check would fail...

forsyth2: They appear identical to me, yes. And `diff.show()` just shows a black image, which would indicate nothing is different between them, unless there is an invisible diff, perhaps one pixel? I don't know what's happening exactly.

chengzhuzhang: I would suggest to figure out this problem before moving forward. After all, we have largest number of images in lat_lon. And would expect the image checker to return credible results...

forsyth2: Alright, I'll do some more investigation. `test_lat_lon_regional` does pass consistently at least.

tests/system/test_diags.py
```diff
@@ -65,25 +75,49 @@ def setUpClass(self):
...
+            self.assertIsNone(diff.getbbox())
+        else:
+            # Sometimes, a few pixels will differ, but the two images appear identical.
+            # https://codereview.stackexchange.com/questions/55902/fastest-way-to-count-non-zero-pixels-using-python-and-pillow
+            nonzero_pixels = diff.crop(bbox).point(lambda x: 255 if x else 0).convert("L").point(bool).getdata()
+            num_nonzero_pixels = sum(nonzero_pixels)
+            print('\npath_to_actual_png={}'.format(path_to_actual_png))
+            print('path_to_expected_png={}'.format(path_to_expected_png))
+            print('diff has {} nonzero pixels.'.format(num_nonzero_pixels))
+            self.assertTrue(num_nonzero_pixels < 5)
```
forsyth2, line 119: @chengzhuzhang For some reason, the `lat_lon` actual and expected plots will differ by as many as 4 pixels. My guess is that the plotting software doesn't always draw the contours identically each time. Luckily, a difference this small doesn't appear to actually be visible. Therefore, I believe this check on the number of nonzero pixels is sufficient. Would you agree?

chengzhuzhang: Thanks for investigating. Yes, good idea to have this additional check. I think we can proceed with this (<5) threshold and we can refine if needed after trying this on more plots.

PR #368, opened 2020-12-12
Move testing data

forsyth2:
Move testing data out of the repo. Resolves #303. Data will be moved to a directory on the E3SM Data Server on Blues/Anvil. The tests will automatically copy down the data from that server to run.

Note that the `tests/climo` directory is currently unused. Also note that this pull request simply removes the `.nc` files from the repo; #306 (removing `.nc` files from the entire git history) will still need to be addressed.

PR #372, opened 2020-12-23
Script to download test data

forsyth2:
Download test data from https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/ rather than using `scp -r`, as discussed in #368.

### DIFF
--- a/tests/download_data.py
+++ b/tests/download_data.py
@@ -1,0 +1,74 @@
+import os
+import re
+import urllib.request
+
+
+# https://stackoverflow.com/questions/49113616/how-to-download-file-using-python
+def retrieve_file(url, file_path):
+    dir_path = os.path.join(*os.path.split(file_path)[:-1])
+    # https://stackoverflow.com/questions/12517451/automatically-creating-directories-with-file-output
+    if dir_path:
+        os.makedirs(dir_path, exist_ok=True)
+    print('Downloading {} to {}'.format(url, file_path))
+    urllib.request.urlretrieve(url, file_path)
+    return file_path
+
+
+def download_files(url_prefix, url_suffix, directory_prefix=None):
+    print('url_prefix={}'.format(url_prefix))
+    print('url_suffix={}'.format(url_suffix))
+    print('(local) directory_prefix={}'.format(directory_prefix))
+    url = os.path.join(url_prefix, url_suffix)
+    if directory_prefix:
+        links_file_path = os.path.join(directory_prefix, url_suffix)
+    else:
+        links_file_path = url_suffix
+    links_file_path = '{}.html'.format(links_file_path)
+    print('Downloading files from {}; checking for links on {}'.format(url, links_file_path))
+    html_path = retrieve_file(url, links_file_path)
+    links = []
+    with open(html_path, 'r') as html:
+        for line in html:
+            match_object = re.search('<a href="(.*)">', line)
+            if match_object:
+                link = match_object.group(1)
+                # Ignore '../'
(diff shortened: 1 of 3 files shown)


forsyth2 (commented):
@chengzhuzhang This pull request will allow the automated tests to download the test data without needing a SSH connection to `Blues`. It does have the disadvantage that all files to be downloaded must be manually listed in the script. (`scp -r` downloaded the complete directories)

However, it looks to me like https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/download_data.py#L55 also doesn't really get around that problem; it lists the necessary files in a text file.

forsyth2:
I'm looking for a way to download all the files from the web server without explicitly listing them. It would be unreasonable to list every image for the full set run (https://portal.nersc.gov/cfs/e3sm/zhang40/tutorials/run_v230_allsets/viewer/).

forsyth2:
@chengzhuzhang This is ready for review. The download script is convoluted -- it goes through the web server finding files to download and subdirectories to go into. This takes about 4 minutes (`scp` took about 2 minutes) but 1) no login to Blues/Anvil is required and 2) there is no need to specifically list which files to download. Running the tests, with this download step, takes about 6 minutes.

chengzhuzhang (approved):
This looks good!

chengzhuzhang:
@forsyth2 The Travis CI test failed though, it seems like one of the image checks failed?

forsyth2:
@chengzhuzhang That's strange. I just reran the tests twice locally and they pass. I wonder if something is different in the Travis build -- it looks like the polar image is off by more than five pixels. I wonder if we should increase the max number of mismatched pixels?

forsyth2:
@chengzhuzhang Actually, I just noticed the test prints the number of nonzero pixels -- 189! I don't understand how that can be so different from the local run.

forsyth2:
I ran the Travis CI test again and it came back with the same number of mismatched pixels -- 189. https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/unit_test_images/polar/ERA-Interim/ERA-Interim-T-850-ANN-polar_S.png is the image causing problems.

chengzhuzhang:
It is quite some difference. This is puzzling. I recall that we had faulty polar plots at one time when `cartopy` was updated. But it does seem the Travis CI build gets the good version of `cartopy`. I'm wondering if we can try to `download` the image generated from Travis and to visual check?

forsyth2:
@chengzhuzhang I can't find a simple way of downloading the image generated by Travis CI. (It looks like I would have to set up some sort of deployment?) I did however discover that running on Cori in the `e3sm_unified_1.4.0` environment _also_ causes a mismatch of 189 pixels on that image. The diff of the image produced on Cori and the image produced locally is blank as far as I can tell. Apparently, 189 pixels isn't enough to be visible.

forsyth2:
@chengzhuzhang The images have 2,103,750 pixels. 189 mismatched pixels / 2,103,750 total pixels = 0.00898% mismatch rate. Do you think I should just change the mismatched pixel check to use a percentage rather than a flat number?

chengzhuzhang:
Good idea!

forsyth2:
The check on fraction of mismatched pixels worked.

Going forward, all unit tests should pass on Travis CI before merging code into `master`.

forsyth2:
Confirming that tests on `master` pass: https://travis-ci.org/github/E3SM-Project/e3sm_diags/builds/753451220

PR #385, opened 2021-02-09
Integrate GitHub Actions CI/CD workflow

tomvothecoder:
## Description

This PR replaces the TravisCI and CircleCI services with GitHub Actions. 

Why?
- TravisCI is no longer providing CI minutes to open source/free-tier projects. Once the trial runs out, we are forced to upgrade.
- CircleCI was not configured correctly and originally only used for testing on Mac OS.
- GitHub Actions integration means CI is housed under the GitHub ecosystem, therefore simplifying the software stack. 

Summary of Changes
- Remove TravisCI and CircleCI config files
- Add GitHub Actions CI/CD workflow
  - Runs pre-commit hooks with linters and formatters
  - Builds Conda env and run test suite on Linux and MacOS
  - Publishes to Anaconda when test suite passes on push to master
- Add GitHub Actions badge to README.md
- Run pre-commit hooks on existing files
  - `trailing-whitespace`, `end-of-file-fixer`, `check-yaml` 

Fixes # (issue)
- Closes #384 

## Type of change

- New feature (non-breaking change which adds functionality)

PR #389, opened 2021-02-23
Docs for testing

### LINKED ISSUE #386: Add developer docs for updating expected images
Add documentation (for developers) to explain how to properly update expected images for the complete run image comparison test.

forsyth2:
Add documentation for testing E3SM Diags. Resolves #386.


source/dev_guide/testing.rst
```diff
@@ -0,0 +1,73 @@
...
+    .. code::
+
+        pip install . # Install your changes
+        ./tests/test.sh # Run all unit tests
+
+Automated tests
+---------------
+
+We have a Continuous Integration / Continuous Delivery (CI/CD) workflow.
+The unit tests are run automatically as part of this.
```
forsyth2, line 18: @tomvothecoder I figure you can update this part once #385 merges.

source/dev_guide/testing.rst
```diff
@@ -0,0 +1,73 @@
...
+If this test passes, there is no need to continue. If it fails however, that means
+your code has changed what the output looks like.
+
+If you expected your code to do this, then you need to update the expected images.
+
+    .. code::
+
+
+        ls image_check_failures # These are the images that differ from expected.
+        # <date> below should be in the form `yyyy_mm_dd`.
```
forsyth2, line 57: @tomvothecoder Would it be better to use the commit hash (i.e., the hash of the latest commit on a developer's pull request) than the date? That way we can see what commit produced a certain set of expected images. However, the `<date>` approach is much easier to sort and understand.

Or is it even worth keeping older versions? If we don't keep older versions we can keep the paths set to one expected images directory, no edits required. However, it may be useful to keep a record of what the images looked like on previous runs.

forsyth2 (commented):
@chengzhuzhang @tomvothecoder This is ready for review. As with other docs changes, you only need to review the changes in `source/`. I added a couple comments but please let me know if 1) the guide doesn't make sense or 2) you think the testing workflow should change.

source/dev_guide/testing.rst
```diff
@@ -0,0 +1,73 @@
...
+    .. warning::
+        You have to run this test manually. It is not run as part of the CI/CD workflow.
+
+If you've been developing code on a different machine, you can get the code to a LCRC
+machine by doing the following. First, push your code to GitHub.
+Then log into a LCRC machine and do the following:
+
+    .. code::
+
+        git fetch <repo-name> <branch-name> # Fetch the branch you just pushed
```
forsyth2, line 37: `<fork-name>` might be better than `<repo-name>` here.

chengzhuzhang:
@forsyth2 I like the idea of versioning the expected image directories. We could have a directory with name `expected_images_latest`. If there is update to the directory, the updated one will be renamed as latest, while time stamp the old version. Somehow a relevant point, for each release, it would be nice if we can set a designated directory to save the viewer from a complete run, so changes from version to version can be nicely captured.

source/dev_guide/testing.rst
tomvothecoder: Makes sense, will do after merging that PR

source/dev_guide/testing.rst
tomvothecoder: I agree since we use the forking model and should run the tests against the fork

forsyth2:
@chengzhuzhang It would be cleaner to put the older versions in their own directory rather than the latest ones.

I can do that, but the issue I had with using the suffix `_latest` is we'd have to remember what date that was updated. `ls -l` showing the month/day might be sufficient since we'd expect to change the expected images more than once per year (e.g., with at least each release).

source/dev_guide/testing.rst
tomvothecoder: If we only use the `<date>` format, would there be naming collisions with previous runs that overwrite the contents of the folder? 
For example, if you run the test suite on a previous commit and on a new commit, would the generated files from the previous commit be overwritten by the new commit? If it does get overwritten, I might suggest using a timestamp  (e.g., `yyyy_mm_dd_hh_mm`) and/or hash of the commit that the test suite was on. Timestamp would allow uniqueness so you can keep previous runs on the same branch, and the hash of the commit can make it clearer which commit the test suite was run on.

I think it would be worth keeping older versions of test suites to see if there is any regression between commits and their associated code changes. This works hand-in-hand with labeling the commit hash mentioned above.

tomvothecoder (approved):
Looks good to me

chengzhuzhang:
@forsyth2 Yeah, I think `ls -l` should be sufficient to show the dates for `_latest`.  We can autogenerate the timestamp and attach to directory name, so that it don't need to be changed each time

source/dev_guide/testing.rst
forsyth2: @tomvothecoder Well if you run the complete-run test, nothing is automatically transferred to the expected images, so there'd be no automatic over-writing.

Maybe `yyyy_mm_dd_hash` would be good? But hashes are long -- maybe we could just list the first 7 characters like on https://github.com/E3SM-Project/e3sm_diags/commits/master?

@chengzhuzhang If we do `yyyy_mm_dd_hash`, I think just keeping the path set to `_latest` would be difficult, because then people would have to manually keep track of both the date and commit associated with the latest directory.

forsyth2: I think `yyyy_mm_dd_<first 7 characters of hash>` makes the most sense as a suffix. I actually see the need to edit the expected_file paths as a positive in that it makes it clear in the pull request that this code changes the expected output.

If we're good with that, I will edit the docs accordingly and merge this.

chengzhuzhang: We may think about the workflow more. Something in my mind
```
generate `complete_run_latest.py`, point to `complete_run_v2.3.py`
If `complete_run_latest.py` need to be updated (i.e., new feature introduced):
     generate  new `complete_run_v.x.x.py`, `complete_run_latest.py`, point to `complete_run_vx.x.py`
     Create result directory `actual`_timestamp
             if the changes are expected: 
                  point `expected`_latest to `actual`_timestamp

if `complete_run_v2.3.py` stays the same (i.e., bug fixes or small changes):
     Create result directory `actual`_timestamp
           fix any unexpected diff. 
```
Something like this: a simple system to track change of the complete_run.py script and the results directory. We can add hash to the file/directory names if that's preferred.

forsyth2: `complete_run_v2_3_0.py` isn't really being changed all that much, except for those paths. (And I suppose for each new release, we'll have a new `tests/complete_run_vMAJOR_MINOR_PATCH.py` and `examples/tutorial/tutorial_vMAJOR_MINOR_PATCH_all_sets_E3SM_machines.py`.) 

Perhaps `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/` could have the subdirectories `previous_versions` and `latest`. `latest` could have a README specifying hash and date that `expected_images_all_sets.txt` and `all_sets` are from. Whenever `latest` is updated, developers will move the old ones into `previous_versions`.

chengzhuzhang: This sounds good. Maybe it is also good to add something like: copy the complete run of a new release to a new directory under /lcrc/group/e3sm/public_html/e3sm_diags_test_data/viewers/ with a standard filename (`_vMAJOR_MINOR_PATCH` as suffix)?

forsyth2:
@chengzhuzhang @tomvothecoder I created a second commit to address your suggestions. Let me know if those changes (and the suggested workflow) look good.

These changes are in tandem with #391 which edits `master` instead of `gh-pages`. Please review those changes too.

For that pull request, running `./tests/test.sh` passes. Running `python -m unittest complete_run.py` fails however. Two Taylor diagrams differ from expected -- see https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.forsyth2/image_check_failures_2021_02_23/. It looks like `global_ceres_ebaf_toa` got `v4.1` added on top of `v4.0`. They're haven't been any new commits added on `master` since we last ran this test however. Is it possible that adding `v4.1` to `/lcrc/group/e3sm/public_html/diagnostics/observations/Atm/climatology/` caused this change? `ls -l` shows it being added on 2020-02-14 -- over a year ago.

chengzhuzhang:
@forsyth2 I did rsync last night to update the observation directory. Most likely this action cased the updated Taylor Diagram. We can use the new directory as a new baseline for images in this case.

forsyth2:
@chengzhuzhang @tomvothecoder This is ready for code review. There are also minor changes to `master` in #391 (`./tests/test.sh` and `python -m unittest tests/complete_run.py` pass). Please review that in tandem with this. If the workflow looks good, I'll merge these pull requests.

docs/html/_sources/dev_guide/testing.rst.txt
```diff
@@ -0,0 +1,86 @@
+Testing E3SM Diagnostics
+==============================================
+
```
tomvothecoder, line 3: I will add an anchor link here to set up the developer environment on #390.

Something like "Make sure to have your developer environment setup locally and on the test server."

source/dev_guide/testing.rst
```diff
@@ -0,0 +1,86 @@
+Testing E3SM Diagnostics
+==============================================
+
```
tomvothecoder, line 3: I will add an anchor link here to set up the developer environment on #390.

Something like "Make sure to have your conda developer environment setup locally and on the test server."

source/dev_guide/testing.rst
```diff
@@ -0,0 +1,86 @@
...
+    .. code::
+
+        pip install . # Install your changes
+        ./tests/test.sh # Run all unit tests
+
+Automated tests
+---------------
+
+We have a Continuous Integration / Continuous Delivery (CI/CD) workflow.
```
tomvothecoder, line 17: I will add to this section for GitHub Actions on #390

tomvothecoder (approved):
Your latest changes look good to me

chengzhuzhang (approved):

Issue #393, opened 2021-02-25
Update conda dev environment dependencies


tomvothecoder:
Building a conda environment with `e3sm_diags_env_dev.yml` in its current state on Linux machines causes the following issue:

![image](https://user-images.githubusercontent.com/25624127/109083853-8b1fa180-76bb-11eb-8659-a0cb6210e1db.png)

There seems to be mismatching dependencies resulting in conflicts.

The solution would be to create a conda environment with the base dependencies for the dev environment, then export it to the respective .yml file.

EDIT: New proposed solution https://github.com/E3SM-Project/e3sm_diags/pull/394#issuecomment-786020056

xylar:
I know you all do like having the yaml files for your development environments but I think this is a really fragile way of defining things.  For MPAS-Analysis, we have a development environment just defined as the command for creating the env:
```
conda config --add channels conda-forge
conda config --set channel_priority strict
conda create -n mpas-analysis python=3.8 numpy scipy "matplotlib>=3.0.2" \
    netCDF4 "xarray>=0.14.1" dask bottleneck lxml "nco>=4.8.1" pyproj \
    pillow cmocean progressbar2 requests setuptools shapely "cartopy>=0.18.0" \
    cartopy_offlinedata "geometric_features>=0.1.13" gsw "pyremap<0.1.0" \
    "mpas_tools>=0.0.15" pandas python-dateutil six
conda activate mpas-analysis
```
This way, you're only constraining some relevant packages rather than pinning every single one.

If you really want to preserve an environment that others should use exactly as it is, the only real way to do that is some form of tarring up the packages, e.g.:
https://conda.github.io/conda-pack/
But I don't think that's an appropriate approach for a development environment.  That would be more for preserving a production environment so you could reproduce the results exactly at a later point.

tomvothecoder:
@xylar Thanks for looking into this.

I actually suggested a solution for pinning only the base dependencies for prod and dev environments, which reduces the length and complexity of the yml files and the need to split files by OS. As you mentioned, pinning every single dependency has a high chance of failure and isn't necessary. 

Here's my solution for reference.
https://github.com/E3SM-Project/e3sm_diags/pull/394#issuecomment-786020056

xylar:
I agree, that's going to make your lives a lot easier.

Issue #390, opened 2021-02-23
Update docs to cover QA tools and simplified conda environments 


tomvothecoder:
With PR #381 merged, the docs should be updated to reflect the latest QA tool integrations and how to use them.

Additional sections to cover
- Update Development Environment process (#394)
- Version Control Workflow (now with pre-commit)
- Documentation

PR #394, opened 2021-02-25
Update dev dependencies to resolve pkg conflicts

tomvothecoder:
This PR updates `e3sm_diags_env_dev.yml` and `e3sm_diags_env_dev_osx.yml` to the latest dependencies to resolve package conflicts. Each environment successfully builds on their respective OS now, including the Linux servers.

I ran the command below to get the latest base dependencies for each environment:
```
conda create -y -n <e3sm_diags_env_dev | e3sm_diags_env_dev_osx> -c conda-forge -c defaults python=3.7 cdp cdms2 cdutil genutil cdtime numpy matplotlib cartopy beautifulsoup4 lxml black isort flake8 flake8-isort pre-commit
```

Then I exported the environment:
```
conda env export --no-build> <e3sm_diags_env_dev.yml | e3sm_diags_env_dev_osx.yml>
```

On Linux servers use:
```
conda env create -n e3sm_diags_env_dev -f e3sm_diags_env_dev.yml
```


chengzhuzhang (commented):
Hi Tom, 
I'm able to create development env without conflicts. My only suggestion is to also update dependencies in e3sm_diags_env(env_osx).yml in this PR. Based on the development env yamls, you will just need to:
* add `e3sm` under `default` in channel session
* add `e3sm=2.4.0` as a dependency
* edit `name` (first line) and `prefix` (last line)

tomvothecoder:
> Hi Tom,
> I'm able to create development env without conflicts. My only suggestion is to also update dependencies in e3sm_diags_env(env_osx).yml in this PR. Based on the development env yamls, you will just need to:
> 
> * add `e3sm` under `default` in channel session
> * add `e3sm=2.4.0` as a dependency
> * edit `name` (first line) and `prefix` (last line)

Hey Jill,

I think there is a way to simplify the `.yml` files even further. Since we only know the base dependencies for the production and dev environments, we just need to pin those and let Conda resolve the rest of the dependencies. It will reduce the length and complexity of the `.yml` files significantly.

https://stackoverflow.com/a/39299669

> This answer is given with the assumption that you would like to make sure that the same versions of the packages that you generally care about are on different platforms and that you don't care about the exact same versions of all packages in the entire dependency tree. If you are trying to install the exact same version of all packages in your entire dependency tree that has a high likelihood of failure since some conda packages have different dependencies for osx/win/linux. For example, the recipe for otrobopt will install different packages on win vs. osx/linux, so the environment list would be different.

> Recommendation: manually create an environment.yaml file and specify or pin only the dependencies that you care about. Let the conda solver do the rest. Probably worth noting is that conda-env (the tool that you use to manage conda environments) explicitly recommends that you "Always create your environment.yml file by hand."

chengzhuzhang:
Hey Tom,  This is great. We tried different methods earlier for a more stable installation instruction. But back then i think one of our base dependency was having some problems. But it seems the problem was solved recent years. So we can simplify.To not split cross platforms and to simplify the .yml files is great. Let's use this proposed method to create .yml. Would you go ahead and make the change. I will test the new files. Also need to make a note to update the documentation as well.

chengzhuzhang:
Oh, I just saw @xylar 's comment here: https://github.com/E3SM-Project/e3sm_diags/issues/393#issuecomment-785706935
If i understand correctly, the method Xylar has for Mpas-analysi should mostly equivalent to use a simplified yaml file?

tomvothecoder:
@chengzhuzhang  Great, I will move forward with the proposed solution.

I just saw his comment also. Correct, it is equivalent to using a simplified dev yml file that I suggested, except he is listing the dependencies manually in the conda create command.

tangq:
I just tried to install the development env on my mac laptop with https://raw.githubusercontent.com/E3SM-Project/e3sm_diags/v2.3.0/conda/e3sm_diags_env_dev.yml and got errors (below). The same yml file works on cori. I would be happy to try the simplified version once available.

```
conda env create -n e3sm_diags_env_dev -f e3sm_diags_env_dev.yml
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - nss=3.59
  - libgfortran-ng=9.3.0
  - libxkbcommon=0.10.0
  - libgcc-ng=9.3.0
  - _openmp_mutex=4.5
  - libstdcxx-ng=9.3.0
  - libgomp=9.3.0

```

tomvothecoder:
@chengzhuzhang I've simplified the yml files for both environments and the CI build and complete test suite both pass.

The only issue for `e3sm_diags_env.yml` is the `e3sm` package not being found, even with the `e3sm` channel added. This error also occurs on the `master` version of the .yml file.

```
(base) [ac.tvo@chrlogin1]~/e3sm_diags% conda env create -f conda/e3sm_diags_env.yml         
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - e3sm=2.4.0
```

Any thoughts or suggestions?

chengzhuzhang:
Yes, this should work after we release v2.4.0. Perhaps just leave it there and test later after a release?

xylar:
Should be `e3sm_diags`, not `e3sm`, righit?

chengzhuzhang:
> Should be `e3sm_diags`, not `e3sm`, righit?

Yes, as @xylar pointed out.

tomvothecoder:
@xylar Thanks for the catch!

@chengzhuzhang This PR is ready for final review. The applicable sections to the docs will be updated in #390.

conda/e3sm_diags_env.yml
```diff
@@ -4,172 +4,20 @@ channels:
...
+  # Additional
+  - e3sm_diags=2.4.0
   - pip:
-    - cdms2==3.1.5
-    - pyqt5-sip==4.19.18
-    - pyqtchart==5.12
-    - pyqtwebengine==5.12.1
-    - regrid2==3.1.5
-prefix: /export/zhang40/miniconda3/envs/e3sm_diags_env
+      - cdms2==3.1.5
```
xylar, line 22: I believe `cdms2` is being incorrectly identified as coming from `pypi` (via `pip`).  It is related to how the package gets installed in the conda recipe.  But this is probably not a problem because I don't think `cdms2` can be installed via `pip` in any case so if this environment is working, all packages must be coming from `conda-forge`, not `pypi`.

Still, I would suggest removing the `pip` section entirely and seeing if things still work.

conda/e3sm_diags_env_dev.yml
```diff
@@ -3,177 +3,27 @@ channels:
...
+  # Developer Tools
+  - black=20.8b1
+  - flake8=3.8.4
+  - flake8-isort=4.0.0
+  - mypy=0.790
+  - pre-commit=2.10.1
+  # Testing
+  - dask
+  - scipy
   - pip:
```
xylar, line 27: Same comment as above.  `cdms2` shouldn't come from `pip`.

chengzhuzhang (approved):
Looks good!

conda/e3sm_diags_env.yml
tomvothecoder: Nice, that would make the `.yml` even cleaner. Giving that a shot, thanks for the suggestion,

tomvothecoder: Looks like `cdms2` is coming from pypi, but now installing through `conda` instead of `pip`.

Using `conda info` after moving `cdsm2` out of `pip` section:
```
cdms2                     3.1.5                    pypi_0    pypi
```

PR #400, opened 2021-03-10
Fix pre-commit typos and add `--no-verify` flag tip

tomvothecoder:
This PR fixes typos for `pre-commit` installation instructions, and adds a tip for the `--no-verify` flag.


forsyth2 (approved):
Looks good, thanks!

PR #401, opened 2021-03-10
Add ARM Diags plots to documentation

forsyth2:
Add ARM Diags plots to documentation.


forsyth2:
@chengzhuzhang This is ready for review. The only `source` file changed was `source/index.rst`. I added 4 plots for ARM Diags (see attached image). I know you mentioned combining the plots into a single image, but other diags have multiple images too and I think it is cleaner this way. If you want I can try to make a single image though.

ARM Diags plots 2021_03_09

chengzhuzhang:
@forsyth2 Thank you for taking care of this, for diurnal cycle of cloud, would you replace with observed (reference) plot, which is a better demo plot. In addition, the title of last figure should be "convection onset". Otherwise, the new plots are good to go.

PR #414, opened 2021-03-24
Scale data to minimum number of months

### LINKED ISSUE #412: Allow QBO Diags to work with insufficient ref data
Currently, QBO diags doesn't work if the ref data and test data aren't for the same amount of time. For instance, on https://github.com/E3SM-Project/zppy/pull/2 when 50 years of test data are used, QBO diags fail because there is only 38 years of ref data. We could simply allow the ref data line to end earlier than the test data line in the bottom plot.

forsyth2:
QBO Diags previously failed if test and ref data differed in length. This pull request scales them to be of the same length. Resolves #412

### DIFF
--- a/acme_diags/plot/cartopy/qbo_plot.py
+++ b/acme_diags/plot/cartopy/qbo_plot.py
@@ -91,7 +91,7 @@ def plot(period_new, parameter, test, ref):
 
     fig = plt.figure(figsize=(14, 14))
 
-    months = min(ref["qbo"].shape[0], test["qbo"].shape[0])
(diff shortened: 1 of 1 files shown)


debug.py
```diff
@@ -0,0 +1,39 @@
+import os
```
forsyth2, line 1: Remove this file before merging

forsyth2:
@chengzhuzhang This is ready for review. I tested on Blues/Anvil and the QBO diags worked despite having 20 years of test data and 10 years of ref data. All the unit tests also pass.

chengzhuzhang (approved):
Looks good!

PR #425, opened 2021-03-26
Fail if any part fails

forsyth2:
Currently, E3SM Diags exits successfully if the viewer is generated, not if all sets are successfully included. This pull request makes it so that E3SM Diags fails unless every part succeeds.

### DIFF
--- a/acme_diags/acme_diags_driver.py
+++ b/acme_diags/acme_diags_driver.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from __future__ import print_function
-
+from typing import Dict, Tuple
 import os
 
 # Must be done before any CDAT library is called.
@@ -300,10 +300,22 @@ def run_diag(parameters):
     return results
 
 
+def create_parameter_dict(parameters):
+    d: Dict[type, int] = dict()
+    for parameter in parameters:
+        t = type(parameter)
+        if t in d.keys():
+            d[t] += 1
+        else:
+            d[t] = 1
+    return d
+
+
 def main(parameters=[]):
     parser = CoreParser()
     if not parameters:
         parameters = get_parameters(parser)
+    expected_parameters = create_parameter_dict(parameters)
 
     # each case id (aka, variable) has a set of parameters specified.
     # print out the parameters for each variable for trouble shootting
@@ -350,6 +362,17 @@ def main(parameters=[]):
 
             index_path = create_viewer(path, parameters)
             print("Viewer HTML generated at {}".format(index_path))
+
+    actual_parameters = create_parameter_dict(parameters)
+    if parameters[0].fail_on_incomplete and (actual_parameters != expected_parameters):
+        d: Dict[type, Tuple[int, int]] = dict()
+        # Loop through all expected parameter types.
+        for t in expected_parameters.keys():
+            d[t] = (actual_parameters[t], expected_parameters[t])
+        message = "Not all parameters completed successfully. Check output above for errors/exceptions. The following dictionary maps parameter types to their actual and expected numbers: {}".format(
+            d
+        )
+        raise Exception(message)
 
 
 if __name__ == "__main__":
--- a/acme_diags/parameter/core_parameter.py
+++ b/acme_diags/parameter/core_parameter.py
@@ -80,6 +80,7 @@ class CoreParameter(cdp.cdp_parameter.CDPParameter):
         self.granulate = ["variables", "seasons", "plevs", "regions"]
         self.selectors = ["sets", "seasons"]
         self.viewer_descr = {}
+        self.fail_on_incomplete = False
 
     def check_values(self):
         # must_have_params = ['reference_data_path', 'test_data_path', 'results_dir']
--- a/docs/source/available-parameters.rst
+++ b/docs/source/available-parameters.rst
@@ -266,6 +266,7 @@ Other parameters
 -  **granulate**: Default is ``['variables', 'seasons', 'plevs', 'regions']``.
 -  **selectors**: Default is ``['sets', 'seasons']``. See :ref:`Using the selectors parameter <selector-ex>`.
 -  **viewer_descr**: Used to specify values in the viewer. Default ``{}``.
+-  **fail_on_incomplete**: Exit status will reflect failure if any parameter fails to complete. Default is ``False`` (e.g., a failing parameter will not create a failing exit code).
 
 Deprecated parameters
 ~~~~~~~~~~~~~~~~~~~~~


forsyth2:
@golaz This addresses the issue you were running into where QBO Diags would have an error but E3SM Diags would continue on.

It appears to me that this issue only comes up when `multiprocessing = True`. When I was running locally without multiprocessing, it failed immediately after failing on QBO diags.

The way this pull request is set up, all diags will still run, but the final success of the exit code depends on whether or not all diags have finished without error. Specifically, it checks that all the parameters finish successfully. I tested by forcing QBO diags to fail. I don't think this will work in the more general case where one of the sets using `CoreParameter` fails. I can try more expansive testing (e.g., on different diags failing) if we want this to be more robust.

@chengzhuzhang Should I make an optional parameter to fail in this case or is it good to always fails in this case?

golaz (approved):
This looks good to me.

chengzhuzhang (approved):
Hi Ryan, this looks good to me. Just to confirm if i understand correctly: If one of specified sets fails, the other sets still run but the new implementation raises additional error message indicates the failed sets?

forsyth2:
> Hi Ryan, this looks good to me. Just to confirm if i understand correctly: If one of specified sets fails, the other sets still run but the new implementation raises additional error message indicates the failed sets?

@chengzhuzhang Yes, that's correct -- everything still runs and failures are listed at the end. Note that it only lists the number of failed parameters. (It's unfortunately more difficult to automatically determine _which_ parameters fail exactly). Do you want me to make this an optional fail (i.e., user passes in parameter to fail in this case)?

forsyth2:
@chengzhuzhang Actually I want to debug this further. Since there appears to be multiple copies of a `*Parameter` object created, the returned number could be misleading...

chengzhuzhang:
@forsyth2 The print-out is useful for trouble shooting,  but we would want to print the name of the failed set(s), so debugging will be more targeted.

forsyth2:
@chengzhuzhang I think the only way to get the individual stack traces/errors would be to edit https://github.com/CDAT/cdp/blob/master/cdp/cdp_run.py and actually the way that is implemented, I don't know how easy changing it would be. 

I can get the type of the parameter objects that failed. That would let us know if it was `QboParaemter`, `EnsoDiagsParameter`, etc. The one caveat is that `CoreParameter` maps to multiple sets (`lat_lon`, `polar`, and more). So we'd only know that `CoreParameter` failed -- not which set failed. 

That said, once the user is alerted something failed, they can then look through the output for the stack traces themselves. (That's tedious but at least possible).

chengzhuzhang:
All right. It makes sense, we can have more specified failure information at a later time. I would modify the error info to:
raise Exception("{}/{} parameter(s) failed. Check output above.".format(diff, num_expected_parameters)) though. What the numbers (diff, num_expected) do you get with a failed QBO setting?

zshaheen:
> @chengzhuzhang Actually I want to debug this further. Since there appears to be multiple copies of a `*Parameter` object created, the returned number could be misleading...

This is probably due to the `granulate` option, which splits up parameters into multiple smaller parameters. Again, that was done because often, expensive plot sets (like 4) would result in extra time being taken. I.e. the other plot sets would be done, with 4 just remaining.

> That said, once the user is alerted something failed, they can then look through the output for the stack traces themselves. (That's tedious but at least possible).

That CDP run stuff uses Dask. If I remember correctly, when the multiprocessing (and not distributed) feature is used, it just uses Python's `multiprocessing` library. You can easily just not use `cdp_run` and instead use a `multiprocessing.Pool`.

With this, you can share the state between the processes, [as seen here](https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes). This can allow you to write more detailed error messages that can be printed out in the end.

After almost 3.5 (!) years of this current implementation, revisiting it would be a good idea!

acme_diags/driver/cosp_histogram_driver.py
```diff
@@ -120,4 +120,5 @@ def run_diag(parameter):
                     parameter,
                 )
 
+    raise Exception("Fail!")
```
forsyth2, line 123: Remove before merging

examples/tutorials/tutorial_v2_4_0_all_sets_E3SM_machines.py
```diff
@@ -215,7 +215,7 @@ def run_all_sets(html_prefix, d):
     # run_compy('/compyfs/www/<username>/')
 
     # Results will be at https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/<username>/v2_4_0_all_sets/viewer/
-    # run_lcrc('/lcrc/group/e3sm/public_html/diagnostic_output/<username>/')
+    run_lcrc("/lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/")
 
     # Results will be at https://portal.nersc.gov/project/e3sm/<username>/v2_4_0_all_sets/viewer/
-    run_nersc("/global/cfs/cdirs/e3sm/www/<username>/")
+    # run_nersc("/global/cfs/cdirs/e3sm/www/<username>/")
```
forsyth2, line 221: Revert changes in this file before merging.

forsyth2:
@chengzhuzhang I set it up to report which parameters failed (not quite as fine-grained as which sets failed, but it's a good start). See the second commit (https://github.com/E3SM-Project/e3sm_diags/pull/425/commits/28a4e7c9e6be4aac614c745657fd3c8cb24a7cb0). 

For example running the tutorial set and forcing the histogram plots to fail I get:
```
Exception: Not all parameters completed successfully. Check output above for errors/exceptions. The following dictionary maps parameter types to their actual and expected numbers: {<class 'acme_diags.parameter.core_parameter.CoreParameter'>: (548, 562), <class 'acme_diags.parameter.zonal_mean_2d_parameter.ZonalMean2dParameter'>: (16, 16), <class 'acme_diags.parameter.meridional_mean_2d_parameter.MeridionalMean2dParameter'>: (16, 16), <class 'acme_diags.parameter.enso_diags_parameter.EnsoDiagsParameter'>: (9, 12), <class 'acme_diags.parameter.qbo_parameter.QboParameter'>: (1, 1), <class 'acme_diags.parameter.area_mean_time_series_parameter.AreaMeanTimeSeriesParameter'>: (10, 10), <class 'acme_diags.parameter.diurnal_cycle_parameter.DiurnalCycleParameter'>: (30, 30), <class 'acme_diags.parameter.streamflow_parameter.StreamflowParameter'>: (1, 1), <class 'acme_diags.parameter.arm_diags_parameter.ARMDiagsParameter'>: (55, 55)}
```

@zshaheen Thanks for the information. I created #426 to implement our own multiprocessing code with finer-grained error reporting.

forsyth2:
@chengzhuzhang This error message alerts me that ENSO Diags isn't completing all runs in the v2.4.0 tutorial. I made #427 to address this.

chengzhuzhang:
> Do you want me to make this an optional fail (i.e., user passes in parameter to fail in this case)?

Yes, this will be nice!

forsyth2:
@chengzhuzhang This is ready for re-review. There is now a `fail_on_incomplete` parameter that defaults to `False`. I ran the unit tests and confirmed they still pass. I also updated the docs to reflect this new parameter. If it is set to `True`, then if not all parameters complete successfully, then the exit code will reflect failure.

chengzhuzhang (approved):
Thank you Ryan. I think this PR is good to merge. It is nice to be able to update docs on main branch at the same time!

PR #457, opened 2021-05-28
v2.5.0rc1 update

chengzhuzhang:


conda/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "e3sm_diags" %}
-{% set version = "2.4.0" %}
+{% set version = "2.5.0rc1" %}
```
tomvothecoder, line 3: I noticed the `source` section is using source from tarball and not source from git.

This means that when we release on GitHub, that release will contain an outdated SHA reference. 
I'm currently testing this out on `zstash` and `zppy`, but using [source from Git](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#source-from-git) should fix this issue.

tomvothecoder (commented):
Looks good, but the `source` section of `meta.yaml` should be updated to use source from Git.

conda/meta.yaml
tomvothecoder: Also the docs show an example using source from Git.
https://e3sm-project.github.io/e3sm_diags/_build/html/master/dev_guide/releasing-e3sm-diags.html

chengzhuzhang: I think at some point, @xylar suggested to use tarball as source. SHA will be then updated after git release:https://e3sm-project.github.io/zstash/_build/html/master/dev_guide/release.html

tomvothecoder: Hmmm, yeah I'm not sure what the benefit is of one over the other.

Using tarbell requires a secondary commit after a release to update the outdated SHA in `meta.yaml`. It also doesn't play nicely with CI/CD, but this won't affect `e3sm_diags` because we won't automate Anaconda publishing to the `e3sm` channel. 

I think we'll be okay with tarbell here.

conda/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
 
 package:
     name: {{ name|lower }}
     version: {{ version }}
 
 source:
-    url: https://github.com/E3SM-Project/e3sm_diags/archive/v{{ version }}.tar.gz
-    sha256: cd4651e4f45a956e5f55555e813b166cd37d04689dbc56fdb2c2b7e2540245c6
+    git_url: git://github.com/E3SM-Project/e3sm_diags
```
tomvothecoder, line 9: I think the url is wrong here actually

tomvothecoder: https://github.com/E3SM-Project/e3sm_diags.git

conda/meta.yaml
tomvothecoder: Looks good with the change! One less step every release

xylar:
> I think at some point, @xylar suggested to use tarball as source. SHA will be then updated after git release:https://e3sm-project.github.io/zstash/_build/html/master/dev_guide/release.html

Yeah, I think I was being over-zealous with that. It's a good practice to check that you're getting the expected release with a checksum but that's not worth bungle up the automation. So the new approach is fine with me.

PR #462, opened 2021-06-02
Patch to turn off image comparison

forsyth2:
Incorporate changes from the patch in https://github.com/conda-forge/e3sm_diags-feedstock/pull/5 to turn off image comparison. Using a different matploblib version can cause enough differences in the plots to cause tests to fail. This change disables the image comparison part of the tests by default.

### DIFF
--- a/tests/system/test_diags.py
+++ b/tests/system/test_diags.py
@@ -169,15 +169,28 @@ class TestAllSets(unittest.TestCase):
         path_to_actual_png = full_png_path
         path_to_expected_png = "../unit_test_images/{}".format(png_path)
 
-        mismatched_images = []  # type: ignore
-        compare_images(
-            self,
-            mismatched_images,
-            image_name,
-            path_to_actual_png,
-            path_to_expected_png,
-        )
-        self.assertEqual(mismatched_images, [])
+        if "CHECK_IMAGES" in os.environ:
+            # Set `export CHECK_IMAGES=True` to do a pixel-by-pixel image comparison check.
+            check_images = os.environ["CHECK_IMAGES"].lower() in [
+                "true",
+                "yes",
+                "t",
+                "y",
+                "1",
+            ]
+        else:
+            check_images = True
+
+        if check_images:
+            mismatched_images = []  # type: ignore
+            compare_images(
+                self,
+                mismatched_images,
+                image_name,
+                path_to_actual_png,
+                path_to_expected_png,
+            )
+            self.assertEqual(mismatched_images, [])
 
     def check_plots_generic(
         self, set_name, case_id, ref_name, variables, region, plev=None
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
@forsyth2  Hey, let's just have that in the feedstock repo. I think it is still good to have the checker on here...

forsyth2:
@chengzhuzhang Sounds good, but perhaps it might be good to add the option to turn image checking on/off? (i.e., I could merge this PR but with the `check_images` option defaulting to `True`).

xylar:
@chengzhuzhang, I don't think we want to intentionally have patches like this in the feedstock repo.  I think you want an environment variable that determines if images get checked or not.  It would be set to one value in the CI here and the other in CI on the feedstock.  But there would not be a patch.  So you just need to replace the function argument here with an environment variable.  If the variable isn't present, you'll need to decide what the default should be.  I'm fine either way.

chengzhuzhang:
@xylar @forsyth2 Thank you, both!  Having an environment variable to switch the image checker one and off seems to be the best approach here.

tests/system/test_diags.py
```diff
@@ -169,15 +169,22 @@ def check_html_image(self, html_path, png_path, full_png_path):
...
-            image_name,
-            path_to_actual_png,
-            path_to_expected_png,
-        )
-        self.assertEqual(mismatched_images, [])
+        try:
+            # Set `export CHECK_IMAGES=True` to do a pixel-by-pixel image comparison check.
+            check_images = bool(os.environ["CHECK_IMAGES"])
+        except KeyError:
+            check_images = False
```
forsyth2, line 176: @chengzhuzhang I set the default value as `False` because I believe we then wouldn't have to update the feedstock to set the value to `False`. Is that correct, @xylar? However, setting it to `True` would make it easier for testers because the image checks would automatically be run.

Once we decide on the default value I can 1) squash the commits and merge, 2) make a version update commit, 3) release `rc2`.

forsyth2:
#463 (update the version number) is ready to merge once this pull request merges.

tests/system/test_diags.py
chengzhuzhang: I haven't tested the change, but the PR looks fine. Just need to remember not to include `export CHECK_IMAGES=True` in CI for feedstock.

xylar, line 176: A suggestion based on https://stackoverflow.com/a/65407083/7728169

```suggestion
        if "CHECK_IMAGES" in os.environ:
            # Set `export CHECK_IMAGES=True` to do a pixel-by-pixel image comparison check.
            check_images = os.environ["CHECK_IMAGES"].lower() in ['true', 'yes', 't', 'y' '1']
        else:
            check_images = True
```

xylar: I agree that that `True` is the better default.  We can set `CHECK_IMAGES=False` in the feedstock.

xylar: It's generally better to do and `if ... in ..` than a `try ... except` in case where the former is a possibility.  It's both a bit easier to read and more specific.

PR #463, opened 2021-06-02
v2.5.0rc2 update

forsyth2:
Update to `v2.5.0rc2`. #462 will be merged, then this will be merged, then https://github.com/E3SM-Project/e3sm_diags/releases will be updated to include `v2.5.0rc2`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 FROM continuumio/miniconda3:4.6.14
 
 LABEL maintainer="zhang40@llnl.gov"
-LABEL version="2.5.0rc1"
+LABEL version="2.5.0rc2"
 
 # Copy the entire project dir because we'll install from source.
 COPY . .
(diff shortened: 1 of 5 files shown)


chengzhuzhang:
Hey @forsyth2 , please do the `rc2` update after #464. Thank you.

forsyth2:
We may try to address #465 before merging this.

xylar:
> We may try to address #465 before merging this.

Yes, I would really prefer to have that fixed as part of the next RC is possible.  Then, we can start testing if `mpi4py` can be built with system MPI, and if so whether ILAMB works with that build.

chengzhuzhang:
> > We may try to address #465 before merging this.
> 
> Yes, I would really prefer to have that fixed as part of the next RC is possible. Then, we can start testing if `mpi4py` can be built with system MPI, and if so whether ILAMB works with that build.

Hey @xylar, Tom has started on this. Is early next week Okay for next RC? It would be really nice to have e3sm-diags and ilamb to co-exist in one env.

xylar:
Yep, no problem!

PR #466, opened 2021-06-03
Stop tests if unit tests fail

forsyth2:
Currently, if a unit test fails, then the integration tests still run, meaning the return value of running the test script is still 0 (i.e., success). This change will fail the script if the unit tests fail.

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -1,5 +1,7 @@
 # Run this script from the top level directory of the repository.
 
+set -e # Fail if any line fails
+
 # Uncomment if you want to remove the results directory from the previous test run.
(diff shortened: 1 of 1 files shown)


forsyth2:
@tomvothecoder @xylar Should we merge this before #463 is merged / `rc2` is released?

tomvothecoder:
@forsyth2 I don't see a problem with merging it before #463 since the build is still passing

xylar:
@forsyth2, I agree with @tomvothecoder, this is a nice addition that should go in before rc2.  It's quite important to notice in this way if tests are failing.

PR #467, opened 2021-06-05
Helper scripts  to extract data at ARM sites

chengzhuzhang:
High frequency data are available from E3SM history files globally or from limited region (single point output). This PR is to add scripts to:
1.  extract single point data from global history files and reformat to per-variable, per-site .nc files: postprocessing_E3SM_data_for_single_sites_extraction.py 
2. reformat limited region output  to per-variable, per-site .nc files: postprocessing_E3SM_data_for_single_sites.py
to  prepare data for ARM sites diagnostics
Resolve #404

PR #468, opened 2021-06-07
Add CDMS_NO_MPI env variable

### LINKED ISSUE #324: MPI_init_thread erorr when running e3sm_diags within e3sm_unified 1.3.1.1 on cori knl and haswell
MPI_init_thread erorr when running e3sm_diags within e3sm_unified 1.3.1.1 on cori knl and haswell. Error message as below:
```
(e3sm_unified_1.3.1.1) chengzhu@nid00941:~/acme_diags_parameters/acme_diags> python tutorial_2020_climo_sets.py 
*** An error occurred in MPI_Init_thread
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[nid00941:57912] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```
It was okay running on cori login. Also running standalone e3sm_diags env is also fine.

### LINKED ISSUE #465: Set `CDMS_NO_MPI=True`
For e3sm_diags to work right in environments with `mpi4py` included, we need to wait for https://github.com/conda-forge/cdms2-feedstock/pull/68 to go in, then set `CDMS_NO_MPI=True`.  We could do that within e3sm_diags before all `import cdms` calls or we could require it to be set externally before calling `e3sm_diags`.  I'm not sure which way is best.  I can set that variable within the E3SM-Unified activation script, but that would still mean users that installed e3sm_diags themselves and also have `mpi4py` would need to know that they, too, need to set this variable.  Since we always want this behavior, it seems cleanest to me if `e3sm_diags` can handle it internally.  At the same time, checking an environment variable before an import is maybe considered sloppy form in python.

tomvothecoder:
This PR adds `os.environ["CDMS_NO_MPI"] = "True"` and existing env vars from `acme_diags_driver.py` to `acme_diags/__init__.py`.

- Closes #465, Closes #324 
- Related to https://github.com/conda-forge/cdms2-feedstock/commit/d82ae7aeb62586f99fb85aee6d4b0a7bfa21ac4c
- Merged in `v2.5.0rc2`

### DIFF
--- a/acme_diags/__init__.py
+++ b/acme_diags/__init__.py
@@ -3,3 +3,15 @@ import sys
 
 __version__ = "v2.5.0rc1"
 INSTALL_PATH = os.path.join(sys.prefix, "share/e3sm_diags/")
+
+# Disable MPI in cdms2, which is not currently supported by E3SM-unified
+os.environ["CDMS_NO_MPI"] = "True"
+# Must be done before any CDAT library is called.
+os.environ["UVCDAT_ANONYMOUS_LOG"] = "no"
+os.environ["CDAT_ANONYMOUS_LOG"] = "no"
+# Needed for when using hdf5 >= 1.10.0,
+# without this, errors are thrown on Edison compute nodes.
+os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
+# Used by numpy, causes too many threads to spawn otherwise.
+os.environ["OPENBLAS_NUM_THREADS"] = "1"
+os.environ["OMP_NUM_THREADS"] = "1"
--- a/acme_diags/acme_diags_driver.py
+++ b/acme_diags/acme_diags_driver.py
@@ -1,24 +1,13 @@
 #!/usr/bin/env python
 from __future__ import print_function
-from typing import Dict, Tuple
-import os
-
-# Must be done before any CDAT library is called.
-os.environ["UVCDAT_ANONYMOUS_LOG"] = "no"
-os.environ["CDAT_ANONYMOUS_LOG"] = "no"
-# Needed for when using hdf5 >= 1.10.0,
-# without this, errors are thrown on Edison compute nodes.
-os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
-# Used by numpy, causes too many threads to spawn otherwise.
-os.environ["OPENBLAS_NUM_THREADS"] = "1"
-os.environ["OMP_NUM_THREADS"] = "1"
 
 import importlib
+import os
 import subprocess
 import sys
 import traceback
+from typing import Dict, Tuple
 
-import cdms2.tvariable
 import cdp.cdp_run
 
 import acme_diags
@@ -27,9 +16,6 @@ from acme_diags.parameter.core_parameter import CoreParameter
 from acme_diags.parser import SET_TO_PARSER
 from acme_diags.parser.core_parser import CoreParser
 from acme_diags.viewer.main import create_viewer
-
-# turn off MPI in cdms2 -- not currently supported by e3sm_diags
-cdms2.tvariable.HAVE_MPI = False
 
 
 def get_default_diags_path(set_name, run_type, print_path=True):
--- a/acme_diags/acme_diags_vars.py
+++ b/acme_diags/acme_diags_vars.py
@@ -10,15 +10,11 @@ import os
 from typing import Any, Dict, List
 
 import cdms2
-import cdms2.tvariable
 
 import acme_diags
 from acme_diags.acme_diags_driver import get_parameters
 from acme_diags.derivations.acme import derived_variables
 from acme_diags.parser.core_parser import CoreParser
-
-# turn off MPI in cdms2 -- not currently supported by e3sm_diags
-cdms2.tvariable.HAVE_MPI = False
 
 
 def main():
--- a/analysis_data_preprocess/create_OAFLUX_climo.py
+++ b/analysis_data_preprocess/create_OAFLUX_climo.py
@@ -295,25 +295,19 @@ for fi in fisc:
             if ji == "shtfl":
                 outputdattable.id = "SHFLX"
                 outputdattable.units = "W m-2"
-                outputdattable.standard_name = (
-                    "surface_upward_sensible_heat_flux"
-                )
+                outputdattable.standard_name = "surface_upward_sensible_heat_flux"
                 # print ji,' modified'
             if ji == "lhtfl":
                 outputdattable.id = "LHFLX"
                 outputdattable.units = "W m-2"
-                outputdattable.standard_name = (
-                    "surface_upward_latent_heat_flux"
-                )
+                outputdattable.standard_name = "surface_upward_latent_heat_flux"
                 # print ji,' modified'
             f_out.write(outputdattable)
 
     time_axis = f_out.getAxis("time")
     bnds_axis = f_out.getAxis("bound")
     climatology = None
-    climo_start = "".join(
-        [startyear, clim_edges[clim_edge_values[f_index, 0]]]
-    )
+    climo_start = "".join([startyear, clim_edges[clim_edge_values[f_index, 0]]])
     climo_end = "".join([endyear, clim_edges[clim_edge_values[f_index, 1]]])
     # climatology_bnds_value={climo_start, climo_end}
 
@@ -402,9 +396,7 @@ for fi in fisc:
         shell=True,
     )
     call(
-        "".join(
-            ["ncatted -a _FillValue,'^bounds',d,, ", output_hostANDfilename]
-        ),
+        "".join(["ncatted -a _FillValue,'^bounds',d,, ", output_hostANDfilename]),
         shell=True,
     )
     call(
--- a/analysis_data_preprocess/create_PminusE_GPCP-OAFlux_timeseries.py
+++ b/analysis_data_preprocess/create_PminusE_GPCP-OAFlux_timeseries.py
@@ -29,9 +29,7 @@ qflux = fin2("QFLX", time=("1979-01-15", "2013-12-15", "ccb"))
 # print(prect.getTime().asComponentTime())
 # print(qflux.getTime().asComponentTime())
 
-qflux_reg = qflux.regrid(
-    prect.getGrid(), regridTool="esmf", regridMethod="linear"
-)
+qflux_reg = qflux.regrid(prect.getGrid(), regridTool="esmf", regridMethod="linear")
 
 PminusE = prect / 24.0 / 3600.0 - qflux_reg  # conver to kg/m2/s
 PminusE.setAxis(0, prect.getTime())
@@ -51,7 +49,9 @@ delattr(PminusE, "dataset")
 delattr(PminusE, "var_desc")
 delattr(PminusE, "valid_range")
 
-output_path = "/p/user_pub/e3sm/zhang40/analysis_data_e3sm_diags/GPCP_OAFLux/time_series/"
+output_path = (
+    "/p/user_pub/e3sm/zhang40/analysis_data_e3sm_diags/GPCP_OAFLux/time_series/"
+)
 output_filename = "PminusE_197901_201312.nc"
 fout = cdms2.open(output_path + output_filename, "w")
 fout.write(PminusE)
--- a/analysis_data_preprocess/fix_units_TRMM-3B43v-7_3hr.py
+++ b/analysis_data_preprocess/fix_units_TRMM-3B43v-7_3hr.py
@@ -2,7 +2,9 @@ from pathlib import Path
 
 import cdms2
 
-datapath = "/p/user_pub/e3sm/zhang40/analysis_data_e3sm_diags/TRMM/climatology_diurnal_cycle/"
+datapath = (
+    "/p/user_pub/e3sm/zhang40/analysis_data_e3sm_diags/TRMM/climatology_diurnal_cycle/"
+)
 for path in Path(datapath).rglob("*.nc"):
     print(path.name)
     print(path)


acme_diags/libs/README.md
```diff
@@ -0,0 +1,13 @@
+# Wrapped Libraries
```
tomvothecoder, line 1: Added a README to the `acme_diags/libs` directory to make its purpose clear and to provide a usage example.

acme_diags/libs/cdms2.py
```diff
@@ -0,0 +1,8 @@
+import os
```
tomvothecoder, line 1: This file will replace all `import cdms2` with `from acme_diags.libs.cdms2 import cdms2`

.pre-commit-config.yaml
```diff
@@ -37,4 +37,4 @@ repos:
     hooks:
       - id: mypy
         args: ["--config=setup.cfg"]
-        exclude: analysis_data_preprocess
+        exclude: (analysis_data_preprocess|examples)
```
tomvothecoder, line 40: Exclude examples directory due to mypy throwing errors related to #467

acme_diags/acme_diags_driver.py
```diff
@@ -28,9 +28,6 @@
 from acme_diags.parser.core_parser import CoreParser
 from acme_diags.viewer.main import create_viewer
 
-# turn off MPI in cdms2 -- not currently supported by e3sm_diags
-cdms2.tvariable.HAVE_MPI = False
```
tomvothecoder, line 32: Setting the `HAVE_MPI` is no longer necessary since `os.environ["CDMS_NO_MPI"]` is now used.

tomvothecoder:
@chengzhuzhang This PR is ready for review and the build is still passing.

acme_diags/libs/README.md
chengzhuzhang: Very nice!

chengzhuzhang (approved):
Thank you Tom. The PR is good. This PR should also resolve https://github.com/E3SM-Project/e3sm_diags/issues/324
Would you try install `mpi4py` in your environment , and run `import cdms2` (without this PR), and `import acme_diags.lib.cdms2` (with this PR) and `import cdms2` to confirm?

xylar:
@chengzhuzhang, I was about to request the same test.  If that works out, I'm guessing we're set!

tomvothecoder:
@chengzhuzhang  I switched to the `master` branch and ran `conda install mpi4py` in my dev environment. Afterwards, I did `python -c import cdms2` but no error appeared.

Any ideas on what I should do next to reproduce the `MPI_init_thread error`?

chengzhuzhang:
Hey Tom, are you testing on macos or linux? The error was discovered on `cori` and with an old version of `cdms2` (3.1.4?) Not sure if both contributed.

tomvothecoder:
@chengzhuzhang I tried it locally and on Chrysalis with `cdms2=3.1.5`. I'll try `cdms2=3.1.4` and on cori.

tomvothecoder:
> @chengzhuzhang I tried it locally and on Chrysalis with `cdms2=3.1.5`. I'll try 1) `cdms2=3.1.4`, 2) cori 3) cori with what you did [here](https://github.com/E3SM-Project/e3sm_diags/issues/324#issuecomment-674481431).

I was unable to reproduce the error on Cori using the latest dev env with different `cdms2` versions. (#1 and #2).

I did reproduce the error using the ad-hoc env from #3 with `cdms2=3.1.5`.
```
conda create -y -n cdms_v821_mpi4py_openmpi_py37 -c cdat/label/v8.2.1 -c conda-forge -c defaults python=3.7 mpi4py cdms2 "libnetcdf=*=mpi_openmpi_*" "esmf=*=mpi_openmpi_*" "esmpy=*=mpi_openmpi_*"
```
```
salloc --nodes=1 --partition=debug --time=00:30:00 -C haswell
conda activate cdms_v821_mpi4py_openmpi_py37
python -c "import cdms2"
```

---
For the `e3sm_diags` dev env, it uses the `nompi` versions of libraries through `cdms2`. This might explain why the dev environment does not throw MPI errors.
```
esmf                      8.1.1           nompi_h3060d44_0    conda-forge
esmpy                     8.1.1           nompi_py37h84cd2ed_0    conda-forge
libnetcdf                 4.8.0           nompi_hcd642e3_103    conda-forge
```

I believe what I have to do is overwrite the `nompi` versions of these libraries, similar to the ad-hoc environment.

chengzhuzhang:
Hey @tomvothecoder, on Cori I did:
```
conda create -y -n cdms2_mpi4py -c conda-forge cdms2 mpi4py
salloc --nodes=1 --partition=debug --time=00:30:00 -C haswell
conda activate cdms2_mpi4py
```
Both: python -c "import cdms2"
And run following script
 ```
import os
os.environ["CDMS_NO_MPI"] = "True"
import cdms2
 ```     
Both gave the `MPI_init_thread` error  
        
I get the same results when using `openmpi` build of libnetcdf, esmf and esmpy
`conda create -y -n cdms2_mpi4py_openmpi -c conda-forge -c defaults python=3.7 mpi4py cdms2 "libnetcdf=*=mpi_openmpi_*" "esmf=*=mpi_openmpi_*" "esmpy=*=mpi_openmpi_*"`

It seems like setting the enviroment variable is not working as expected..
      
      
        <span class="pl-k">import</span> <span class="pl-s1">cdms2</span>

tomvothecoder:
> I believe what I have to do is install the `openmpi` versions of these libraries, similar to the ad-hoc environment and E3SM unified.

I was able to reproduce the `MPI_init_thread` error in the dev env with the above.

Now switching to this branch, I am getting the same error in all of these cases:
```
# 1 -- Expected to throw an error without env var
# ====================================================+
import cdms2 

# 2 -- Not expected to throw error with env var
# ====================================================+
from acme_diags.libs.cdms2 import cdms2

# 3 -- Not expected to throw error with env var
# ====================================================+
import os
os.environ["CDMS_NO_MPI"] = "True"
import cdms2
```

I validated that the dev environment is using `cdms2=3.15` with https://github.com/conda-forge/cdms2-feedstock/commit/d82ae7aeb62586f99fb85aee6d4b0a7bfa21ac4c.

Any ideas why the fix might not be working @xylar @jasonb5?

tomvothecoder:
@chengzhuzhang Just saw your comment, thanks! I also tried just `from cdms2 import tvariable` and it doesn't work either.

xylar:
@tomvothecoder and @chengzhuzhang, the MPI from conda-forge doesn't work on Cori or any others of our HPC machines.  I'm sorry, I should have clarified that you should do this test on a Mac or Linux machine where conda-forge MPI will work.

I believe the `MPI_init_thread` error you're seeing isn't related do e3sm_diags, cdms2 or mpi4py specifically.  I believe you will see it if you try to run anything that uses MPI on Cori.  I specifically install the `*=*=nompi*` variants of all packages in E3SM-Unified to avoid this issue.

To combine cdms2 with mpi4py on Cori and other supported HPC machines, we're going to need to build mpi4py with the system MPI and install it.  That's the plan for the next test environment.

xylar:
See this discussion: https://docs.nersc.gov/development/languages/python/parallel-python/#mpi4py-in-your-custom-conda-environment

xylar:
I did some testing on Cori as follows.

Starting from `master` (not this branch) I created a development environment for e3sm_diags as normal:
```
source ~/miniconda3/etc/profile.d/conda.sh
conda activate
conda env create -f conda/e3sm_diags_env_dev.yml
conda activate e3sm_diags_env_dev
```
Then, I set up the modules the same way as E3SM-Unified 1.5.0rc3
<details>
 <summary>modules</summary>

```
module rm PrgEnv-intel
module rm PrgEnv-cray
module rm PrgEnv-gnu
module rm intel
module rm cce
module rm gcc
module rm cray-parallel-netcdf
module rm cray-hdf5-parallel
module rm pmi
module rm cray-libsci
module rm cray-mpich2
module rm cray-mpich
module rm cray-netcdf
module rm cray-hdf5
module rm cray-netcdf-hdf5parallel
module rm craype-sandybridge
module rm craype-ivybridge
module rm craype
module rm papi
module rm cmake
module rm cray-petsc
module rm esmf
module rm zlib
module rm craype-hugepages2M
module rm darshan
module load craype
module load PrgEnv-intel
module load cray-mpich
module rm craype-mic-knl
module load craype-haswell
module swap cray-mpich cray-mpich/7.7.10
module load PrgEnv-intel/6.0.5
module rm intel
module load intel/19.0.3.199
module swap craype craype/2.6.2
module rm pmi
module load pmi/5.0.14
module rm craype-mic-knl
module load craype-haswell
module rm cray-netcdf-hdf5parallel
module load cray-netcdf-hdf5parallel/4.6.3.2
module load cray-hdf5-parallel/1.10.5.2
module load cray-parallel-netcdf/1.11.1.1
module rm git
module load git
module rm cmake
module load cmake/3.14.4
module load perl5-extras

export HDF5_USE_FILE_LOCKING=FALSE
```
</details>

Then, I installed `mpi4py` with `pip` following https://docs.nersc.gov/development/languages/python/parallel-python/#mpi4py-in-your-custom-conda-environment:
```
MPICC="cc -shared" python -m pip install --no-binary=mpi4py "mpi4py==3.0.3"
```

I manually edited 2 files in cdms2 to print the `mpi_disabled` variable:
```
vim ~/miniconda3/envs/e3sm_diags_env_dev/lib/python3.7/site-packages/cdms2/dataset.py \ 
    ~/miniconda3/envs/e3sm_diags_env_dev/lib/python3.7/site-packages/cdms2/tvariable.py 
```
Then, I logged onto an interactive Haswell node.  I loaded the conda environment again, then loaded all the modules as above. Then, I followed the instructions at https://e3sm-project.github.io/e3sm_diags/_build/html/master/quickguides/quick-guide-cori.html#config-and-run
I was expecting to see problems but everything ran fine.  I saw:
```
mpi_disabled False
mpi_disabled False
```
as expected since I'm using `master` and haven't set `CDMS_NO_MPI`.  So it seems like `e3sm_diags` works with `mpi4py` built with system libraries even if MPI is not disabled in `cdms2`.  My suspicion is that the issue we had before was related to other libraries being installed with conda-forge `mpich` or `openmpi`.  I don't think we can expect things to work on Cori or other HPC with those libraries regardless of the `cdms2` MPI flag.  So it will simply be quite important to document for your users on HPC that they need the `nompi` packages unless they're confident that conda-forge `mpich` or `openmpi` works correctly on their machine.

Just to test, I also set `export CDMS_NO_MPI=True` and reran the e3sm_diags example.  I saw:
```
mpi_disabled True
mpi_disabled True
```
but the results were otherwise identical to my untrained eye.

Finally, in this branch, I ran:
```
unset CDMS_NO_MPI
python -m pip install .
python run_e3sm_diags.py
```
To my surprise, I was still seeing:
```
mpi_disabled False
mpi_disabled False
```
I figured out that this is because the statement:
```
import MV2
```
in `acme_diags/derivations/acme.py` (and later elsewhere) also imports `cdms2`.  If I add an unneeded:
```
from acme_diags.libs.cdms2 import cdms2
```
above
```
import MV2
```
I see the expected:
```
mpi_disabled True
mpi_disabled True
```

There's a lot to unpack here.  First, it isn't clear if we need to disable MPI in `cdms2` at all.  Maybe the problem is from conda-forge libraries built for `mpich` or `openmpi` and as long as we exclude them, things are fine whether we set `CDMS_NO_MPI=True` or no.

Second, it seems like you might need to put:
```
# Disable MPI in cdms2, which is not currently supported by E3SM-unified
os.environ["CDMS_NO_MPI"] = "True"
```
right at the top of the main `__init__.py` because it doesn't seem safe to just look for `import cdms2` calls.  `import MV2` calls, `from cdms2 import ...` calls and perhaps others will also import `cdms2`, potentially before `CDMS_NO_MPI=True` gets set.

It would be really helpful to have a discussion today to get this resolved.  Unfortunately, I have meetings from 5 pm straight through to 11 pm (with only a really short dinner break) so I don't have time to have a video conference or anything.  But maybe we can chat over Slack.  My suggestion would be that you make an `rc2` release now with all the other fixes but without this branch for now.  Then, I can make an E3SM-Unified 1.5.0rc4 that includes `mpi4py` built with system MPI but that doesn't have any conda-forge libraries built for `mpich` or `openmpi`.  We can see how `e3sm_diags` behaves both without the `CDMS_NO_MPI` set and when we manually add `export CDMS_NO_MPI=True` before we run.  Based on what we learn from that experience, we can revisit this PR.  Please feel free to start a Slack discussion so we can get this decided today if possible.

tomvothecoder:
@xylar Thanks for hashing out the problem. Your explanations make sense to me!

> There's a lot to unpack here. First, it isn't clear if we need to disable MPI in cdms2 at all. Maybe the problem is from conda-forge libraries built for mpich or openmpi and as long as we exclude them, things are fine whether we set CDMS_NO_MPI=True or no.

I'm also led to believe this is a possibility through my own investigation. 

> Second, it seems like you might need to put:
> 
> ```# Disable MPI in cdms2, which is not currently supported by E3SM-unified
> os.environ["CDMS_NO_MPI"] = "True"```
> right at the top of the main `__init__.py` because it doesn't seem safe to just look for >`import cdms2` calls. `import MV2` calls, `from cdms2 import ...` calls and perhaps >others will also import `cdms2`, potentially before `CDMS_NO_MPI=True` gets set.

Nice catch with `import MV2` reverting the `CDMS_NO_MPI` env var. I agree, setting `CDMS_NO_MPI` before all imports regardless of the package can eliminate this problem.

---
Also, I will make a Slack discussion to confirm these next steps:
1. Release `e3sm_diags` `v2.5.0rc2` without this PR
2. Test E3SM_unified `v1.5.0rc4`
    - Environment Includes `mpi4py` built with system MPI and excludes conda-forge libraries built with `mpich` or `openmpi` 
3. Test `e3sm_diags` 
    - Without `CDMS_NO_MPI`
    - `export CDMS_NO_MPI=True`
    - `CDMS_NO_MPI` in `__init__.py` (this PR)

chengzhuzhang:
@xylar Excellent detective work! I think the next steps you proposed and @tomvothecoder summarized are reasonable. I have a couple of meetings spreading out today,  a slack chat would be good.

acme_diags/acme_diags_driver.py
```diff
@@ -1,24 +1,13 @@
...
-
-# Must be done before any CDAT library is called.
-os.environ["UVCDAT_ANONYMOUS_LOG"] = "no"
-os.environ["CDAT_ANONYMOUS_LOG"] = "no"
-# Needed for when using hdf5 >= 1.10.0,
-# without this, errors are thrown on Edison compute nodes.
-os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
-# Used by numpy, causes too many threads to spawn otherwise.
-os.environ["OPENBLAS_NUM_THREADS"] = "1"
-os.environ["OMP_NUM_THREADS"] = "1"
```
tomvothecoder, line 14: These env vars are now in `acme_diags/__init__.py`

acme_diags/__init__.py
```diff
@@ -3,3 +3,15 @@
...
+os.environ["CDMS_NO_MPI"] = "True"
+# Must be done before any CDAT library is called.
+os.environ["UVCDAT_ANONYMOUS_LOG"] = "no"
+os.environ["CDAT_ANONYMOUS_LOG"] = "no"
+# Needed for when using hdf5 >= 1.10.0,
+# without this, errors are thrown on Edison compute nodes.
+os.environ["HDF5_USE_FILE_LOCKING"] = "FALSE"
+# Used by numpy, causes too many threads to spawn otherwise.
+os.environ["OPENBLAS_NUM_THREADS"] = "1"
+os.environ["OMP_NUM_THREADS"] = "1"
```
tomvothecoder, line 17: These env variables are set globally upon importing any file or module within `acme_diags`.

Examples
- `import acme_diags`
- `from acme_diags import acme_diags_driver`
- `from acme_diags.acme_diags_driver import get_default_diags_path`

acme_diags/acme_diags_driver.py
tomvothecoder: @chengzhuzhang It looks like we can set these env vars globally before any of their related imports.

Let me know if they should only be set through `acme_diags_driver`.

acme_diags/__init__.py
chengzhuzhang: I think it is cleaner now

xylar:
@tomvothecoder and @chengzhuzhang, thanks for the work on this.  I'm really hoping all this effort pays off.  I think it will.

PR #481, opened 2021-06-24
Add stratosphere and troposphere ozone

chengzhuzhang:
This PR is to inlcude Stratosphere column ozone (SCO) and Troposphere column ozone (TCO) into `lat_lon` and `annual_cycle_zonal_mean` sets for both model vs obs, and model vs model runs. 
preprocessing script to generate climo from OMI/MLS observation is provided. 
Note the SCO obs doesn't have longitude dimension, there is some special handling for this in annual_cycle_zonal_mean driver.

Example model vs obs:https://portal.nersc.gov/cfs/e3sm/zhang40/tests/O3_test1/viewer/
Example model vs model: https://portal.nersc.gov/cfs/e3sm/zhang40/tests/O3_test_m2m/viewer/

resolves #481


chengzhuzhang:
@tangq Hey Qi, please look over the examples and let me know if you find any problems.

tangq:
@chengzhuzhang , thanks for working on this.

I looked at the figures on your links. For the `annual_cycle_zonal_mean` figures, I'd suggest 2 minor changes.
1. Change to the contour without interpolations. I am not sure what the Python option is. For NCL, it's called "RasterFill". You can see the Figure 1 of https://gmd.copernicus.org/articles/14/1219/2021/
2. Change the color map for the difference plot to the one similar to Fig. 1 above as well: Red for positive changes and blue for negative. The current colors are weird for ozone plots.

chengzhuzhang:
@tangq Thanks! These should be straightforward change. I will play with the options.  I edited my last reply by adding the question: were SCO and TCO being added to standard monthly output? It would be nice if we can look at them for every runs.

tangq:
Yes, SCO and TCO are output monthly by default.

forsyth2:
@chengzhuzhang Details on the error you encountered:

I tested `python test_run.py TestRun.test_all_sets_and_all_seasons`.

I think as part of this pull request you can replace this code block
```
        all_season_counts = list(season_counter.values())
        # enso_diags and streamflow only run ANN, no seasons
        # So, reduce the ANN count by the number of times these appear
        all_season_counts[0] -= set_counter["enso_diags"]
        all_season_counts[0] -= set_counter["streamflow"]
        if not all(all_season_counts[0] == count for count in all_season_counts):
            self.fail(
                "Counts for the seasons don't match: {}".format(all_season_counts)
            )
```
with this one
```
        # enso_diags and streamflow only run ANN, no seasons
        # So, reduce the ANN count by the number of times these appear
        season_counter["ANN"] -= set_counter["enso_diags"]
        season_counter["ANN"] -= set_counter["streamflow"]
        season_counter["ANN"] -= 1  # 1 lat_lon set only runs "ANN" (no other seasons)
        if not all(season_counter["ANN"] == count for count in season_counter.values()):
            self.fail(
                "At least one season does not match the count for ANN: {}".format(
                    season_counter
                )
            )
```
Then, that test will pass. My changes 1) make the seasons more apparent and 2) deal with the `lat_lon` default diag you added in this PR that only runs "ANN" (`acme_diags/driver/default_diags/lat_lon_model_vs_obs.cfg`).

chengzhuzhang:
@forsyth2 Thank you, Ryan! This makes sense now. I only added `ANN` in cfg so the number of seasons check didnt' pass. This check is useful! I will fix my code by adding all seasons. I will slightly modifying the print out to : "Counts for the seasons defined in .cfg files don't match". Thank you!

chengzhuzhang:
@tangq With the new commit, i changed to use `pcolor` to replace `contourf`, to get the box plots without interplation. I also changed the colormap for diff plot as follows:
image

chengzhuzhang:
@forsyth2 FYI, Tests passed with all seasons being specified for the new variable.

tangq:
@chengzhuzhang , this looks great. Can you change the diff interval to 2, 5, 10, 15, 20, ... and stop at 40. Also, remove the extra zeros on the color bar. Thanks!

forsyth2:
> Tests passed with all seasons being specified for the new variable.

@chengzhuzhang Awesome. I do think the changes I suggested (i.e., just using `season_counter["ANN"]` rather than `all_season_counts[0]`) would be good for code readability, but I can make a GitHub nice-to-have issue to do that later.

chengzhuzhang:
@forsyth2 Okay, it looks like I'm having more changes, I will get it in for this PR as well.

chengzhuzhang:
@tangq How about this one?
image
Thanks for noticing the trailing '.0'. Probably i should remove them for other plot sets.

tangq (approved):
@chengzhuzhang, looks great! It should be ready to be merged.

chengzhuzhang:
I have tested other annual cycle zonal mean plots and they look okay with the plotting. I'm merging now. 
Note, the new obs are updated on Cori and Compy, will update LCRC and ACME1 once machine/accounts problems are resolved.

tangq:
@chengzhuzhang , will these ozone figures be plotted automatically or need some additional settings?

chengzhuzhang:
It works the same as other plotsets like `lat-lon`, `zonal-mean-2d`...(i.e., run `ncclimo` to get climatology files first.)
For the new set, just need to add `from acme_diags.parameter.annual_cycle_zonal_mean_parameter import ACzonalmeanParameter` into the imports session of the run script and then add `annual_cycle_zonal_mean` in `sets_to_run`.

Here is an example run script:
```
import os
from acme_diags.parameter.core_parameter import CoreParameter
from acme_diags.parameter.annual_cycle_zonal_mean_parameter import ACzonalmeanParameter
from acme_diags.run import runner

param = CoreParameter()

param.test_data_path = '/Users/zhang40/Documents/ACME_simulations/'
param.test_name = '20180215.DECKv1b_H1.ne30_oEC.edison' 
param.short_test_name = 'DECKv1b'
param.reference_data_path = '/Users/zhang40/Documents/ACME_simulations/obs_for_e3sm_diags/climatology'
prefix = '/Users/zhang40/Downloads'
param.results_dir = os.path.join(prefix, 'ac_zonal_mean')
param.multiprocessing = True
param.num_workers = 2

runner.sets_to_run = ['annual_cycle_zonal_mean']
runner.run_diags([param])
```
If `SCO` and `TCO` are available from model output, then the plots will be generated automatically (they are already included in the default .cfg files to specify variables).

tangq:
Hi @chengzhuzhang , is it available on chrysalis with the e3sm_unified tool? Or we need to wait for the next e3sm_unified release?

chengzhuzhang:
Yeah, the new release of `e3sm_unified` is under testing. I will make sure to work with Xylar to get this new feature in for this e3sm_unified release :)

tangq:
Thanks for the clarification. I will use this when it's enabled in the e3sm_unified.

PR #485, opened 2021-07-06
Update `setup.py` `packages` setting to include `acme_diags`

### LINKED ISSUE #484: The python package includes the `tests` directory
The line:
https://github.com/E3SM-Project/e3sm_diags/blob/14d7c969305098f0a3ea582ad6a3395c864ae558/setup.py#L118
seems designed to exclude tests but it doesn't actually have the `tests` directory in its excludes.  As a result, it ends up in the conda package and it conflicts with other packages that make the same mistake (e.g. `zppy` and `zstash`).

Instead, it would be much cleaner and safer to do:
```
packages=find_packages(include=['acme_diags', 'acme_diags.*'])
```

tomvothecoder:
This setting was incorrectly configured to ignore test files, but it was not actually working.

- Closes #484

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,7 @@ setup(
     author_email="zhang40@llnl.gov",
     description="E3SM Diagnostics",
     scripts=["acme_diags/acme_diags_driver.py"],
-    packages=find_packages(exclude=["*.test", "*.test.*", "test.*", "test"]),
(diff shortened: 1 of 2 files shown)


tomvothecoder:
@xylar @chengzhuzhang FYI this should fix #484. Luckily, I only had to update `setup.py`.

Issue #482, opened 2021-06-24
Add units to difference plot


chengzhuzhang:
It appears that units are missing from the third panel difference plot. 
![image](https://user-images.githubusercontent.com/13056557/123333808-12a8e000-d4f7-11eb-890a-7f3caaeee157.png)

PR #486, opened 2021-07-06
add units to diff plots, etc

chengzhuzhang:
Add units to diff plots of lat-lon, zonal-mean-2d, meridional-mean-2d. enso-map, polar plots, and QBO pressure-time contour plots. 

Updated images on lcrc for expected integration tests.
Resolves #486


tomvothecoder (approved):
Looks good. I believe this closes #482.

chengzhuzhang:
Thank you! Good catch.

PR #489, opened 2021-07-14
Update load e3sm unified paths

chengzhuzhang:
Updates activation paths for e3sm-unified v1.5.0

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -14,37 +14,32 @@ The paths to ``e3sm_unified`` activation scripts are machine dependent:
 **Compy**
     ::
 
-     source /share/apps/E3SM/conda_envs/load_latest_e3sm_unified.sh
+     source /share/apps/E3SM/conda_envs/load_latest_e3sm_unified_compy.sh
 
 
 **NERSC**
     ::
 
-     source /global/cfs/cdirs/e3sm/software/anaconda_envs/load_latest_e3sm_unified.sh
-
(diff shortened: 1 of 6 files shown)


docs/source/install.rst
```diff
@@ -14,37 +14,32 @@ The paths to ``e3sm_unified`` activation scripts are machine dependent:
...
     ::
 
-     source /usr/local/e3sm_unified/envs/load_latest_e3sm_unified.sh
-
+     source /usr/local/e3sm_unified/envs/load_latest_e3sm_unified_acme1.sh
 
-**Rhea**
-    ::
 
-     source /ccs/proj/cli900/sw/rhea/e3sm-unified/load_latest_e3sm_unified.sh
```
xylar, line 47: @chengzhuzhang, you may wish to add Andes.  I don't yet have my OLCF account re-activated but when I do I will attempt to install there.   However, I don't yet know what the path will be.  So maybe just leave it out for now and we can add it later.

xylar (approved):
This all looks right to me.

I don't think it needs to be done in this PR, but it will be good to add a note somewhere about the difficulties we've had on Cori-KNL and that we recommend against using e3sm_diags (and other diagnostics tools) there.

chengzhuzhang:
@xylar Thank you for reviewing! And nice catch, I added a warning on using knl. Will add Andes later, I also recently re-activated my ORNL account :)

PR #491, opened 2021-07-16
Fix flipped lat of ozone obs

chengzhuzhang:
@tangq reported that the TCO obs lat-lon plots seems not physical: 
i.e: misses the minimum over Tibet in JJA for the obs, which is unphysical (The model result has it).

It turns out, the latitude got flipped when SCO (without longitude) file being attached to TCO (with longitude) using ncks. I have a workaround to flip the lat back in re-processing files.

No source code change, the updated data are now on compy, 

- [x] update other machines.


chengzhuzhang:
Erroneous plot with latitude flipped in observation:
image

New plot with obs data fixed
image

chengzhuzhang:
obs files updated on all machines. no source code change in /acme_diags. Merge now

PR #507, opened 2021-08-09
Update readme

### LINKED ISSUE #504: Examples docs are out of date
The examples docs refer to Edison.  They have the old E3SM-Unified activation script.

It might also make more sense at this point to have the examples work on LCRC in addition to or instead of Cori.  Queue times are much better and the data would be easier to update and maintain.

chengzhuzhang:
resolve #504


chengzhuzhang:
Thank you for catching this! I have replaced outdated links. We have quick guides to cover more machines. The examples are still limited to use NERSC because it is more accessible. But good idea to have those examples working and documented on LCRC too, but will be another PR for Docs updates.

PR #511, opened 2021-08-10
Automate quick guide generation

### LINKED ISSUE #509: Automate quick guide generation
Move the `python source/quickguides/generate_quick_guides.py # Run if quick-guide files were updated` step on `docs/source/contributing.rst` into the automated documentation building (probably at https://github.com/E3SM-Project/e3sm_diags/blob/master/.github/workflows/build_workflow.yml#L128?)

The correct workflow currently is the following:
1. Modify `docs/source/quickguides/generate_quick_guides.py` and/or `docs/source/quickguides/generate_quick_guides.py`
2. Run `python docs/source/quickguides/generate_quick_guides.py` to generate:
  - `docs/source/quickguides/quick-guide-acme1.rst`
  - `docs/source/quickguides/quick-guide-compy.rst`
  - `docs/source/quickguides/quick-guide-cori.rst`
  - `docs/source/quickguides/quick-guide-lcrc.rst`

However, those four generated files currently exist in the repo. They can be (and have been) edited independently, which disrupts the above workflow. It would be better for those generated files to not exist in the repo and have the automated documentation handle the file generation.

forsyth2:
Set up GitHub Actions to handle quick guide generation. Resolves #509.


forsyth2:
@tomvothecoder Is there a good way to test that GitHub Actions will actually now auto-generate the quick guides?

tomvothecoder:
@forsyth2 A little tedious to check, but not hard.

1. Make a PR to your fork
2. [OPTIONAL] Push a very noticeable edit to a quick guide page.
3. Merge the fork PR
4. Check if the command is running in the `CI/CD Build Workflow / publish-docs (pull_request)` step.
5. [OPTIONAL] Check if the HTML build on your fork's GitHub pages 
   - Revert step 2 if successful
6. Merge this PR if successful

docs/source/quickguides/quick-guide-generic.rst
```diff
@@ -1,4 +1,4 @@
-#expand machine_name# quick guide for running e3sm_diags v2
+#expand machine_name# quick guide for running e3sm_diags v2 TEST
```
forsyth2, line 1: Remove before merging

.github/workflows/build_workflow.yml
```diff
@@ -126,6 +126,8 @@ jobs:
       - name: Build Sphinx Docs
         run: |
           cd docs
+          python source/quickguides/generate_quick_guides.py
```
forsyth2, line 129: @tomvothecoder If I run this command locally and then run `make html`, I can generate the proper documentation, but if I run `sphinx-multiversion source _build/html` instead, none of my changes seem to take effect.

This also seems to be the case in my test https://github.com/forsyth2/e3sm_diags/pull/2. After merging that pull request to my fork's `master`, the `gh-pages` doesn't have the new quick guides (see https://github.com/forsyth2/e3sm_diags/tree/gh-pages/_build/html/master/quickguides)

It doesn't look like Sphinx is finding the generated `rst` files -- do they need to be added to `git` perhaps? (see https://github.com/forsyth2/e3sm_diags/runs/3306348047?check_suite_focus=true#step:6:63)

.github/workflows/build_workflow.yml
```diff
@@ -126,6 +126,8 @@ jobs:
...
+          ls source/quickguides
```
forsyth2, line 130: Remove before merging

.github/workflows/release_workflow.yml
```diff
@@ -43,6 +43,8 @@ jobs:
       - name: Build Sphinx Docs
         run: |
           cd docs
+          python source/quickguides/generate_quick_guides.py
+          ls source/quickguides
```
forsyth2, line 47: Remove before merging

.github/workflows/build_workflow.yml
tomvothecoder: Yes, the quick guide `.rst` files need to be pushed because `sphinx-multiversion` uses those source files within each branch/tag for the HTML builds. You probably need to add a separate step to generate quick guides and push them to the repo before `Build Sphinx Docs`.

Refer to lines below for pushing quick-guide `.rst` files to `master` (the `publish-docs` job should only run on `master`). 
https://github.com/E3SM-Project/e3sm_diags/blob/27a12f93b75052ef2f9b62315e13afb3d034c8c2/.github/workflows/build_workflow.yml#L148-L154

forsyth2: > You probably need to add a separate step to generate quick guides and push them to the repo

Does that mean a pull request to `master` that updates the quick guides will not only include the commit from merging the pull request but also a subsequent GitHub actions commit for the generated `rst` files?

tomvothecoder: Yes, this is one issue related with attempting to generate quick guides dynamically using a separate python script.

The other way is to manually run the python script and push the `.rst` files. This might end up being the better solution if the quick guides aren't updated that much, and attempting to automate it doesn't provide that much value.

forsyth2:
Replacing in favor of #513.

Automating quick guide generation would require GitHub actions to make a commit to `master` to update the machine-specific `.rst` files. Those files would then be in `master` just as they would be if the developer had manually generated the files.

My hope was that the machine-specific `.rst` files wouldn't be stored in the repo but rather would only be generated to update the docs. This unfortunately looks to not be possible.

Issue #506, opened 2021-08-06
complete_run test failing


forsyth2:
The `complete_run` test currently fails.

forsyth2:
From discussion on #495:

> However, I'm now debugging a FileNotFoundError: [Errno 2] No such file or directory: './v2_4_0_all_sets/all_sets_v250_20210608_1878984/cosp_histogram/MODIS-COSP/MODISCOSP-COSP_HISTOGRAM_MODIS-JJA-global.png' error....

@chengzhuzhang The above error seems to be because `all_sets_v250_20210608_1878984` (an exact copy of `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/previous_output/all_sets_v250_20210608_1878984`) is in `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets`. I can't `rm -rf` it though because the permissions are `drwxr-xr-x 17 ac.zhang40  E3SM`

I think we need to add a step to change permissions on the directions to replace the expected images (at https://e3sm-project.github.io/e3sm_diags/_build/html/master/dev_guide/testing.html#complete-run-test). Something like `chmod -R g+wX /lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets`

forsyth2:
Expected images have been updated, corresponding to the pre-release https://github.com/E3SM-Project/e3sm_diags/releases/tag/v2.6.0rc1. `./tests/test.sh` and `python -m unittest tests/complete_run.py` pass now.

PR #514, opened 2021-08-13
Consolidate pre-processing scripts

chengzhuzhang:
This PR only update pre-processing scripts, won't change code under e3sm_diags/
-[x] add add pg2 support for TC pre-processing
-[x] add add pg2 support for climo and time-series generation
-[x] add add pg2 support for extracting ARM sites data
-[x] re-org example directory and create new pre-processing scripts folder
-[x] create e3sm v2 (ne30pg2) test datasets: new v2 testing data available on Cori, Compy, LCRC now
-[x] create run script example for e3sm_diags v2.5.0(latest): Note, for v2.5.0 still use `import acme_diags`,  will change to `import e3sm_diags` when v2.6.0 releases.


chengzhuzhang:
This PR only modify examples/pre-processing/docs. The complete test now will use full v2.5.0 sets.

examples/run_v2_5_0_all_sets_E3SM_machines.py
```diff
@@ -0,0 +1,236 @@
+import os
+
+from acme_diags.parameter.annual_cycle_zonal_mean_parameter import ACzonalmeanParameter
```
forsyth2, line 3: I agree we should use `acme_diags` since that's what is needed for v2.5.0. However, the `complete_run` test, when run in the latest development environment, expects `e3sm_diags`. Maybe the best solution is to have `complete_run` use a copy of this file that trades `acme_diags` for `e3sm_diags`?

examples/run_v2_5_0_all_sets_E3SM_machines.py
```diff
@@ -0,0 +1,236 @@
...
+    )
+
+    return param.results_dir
+
+
+if __name__ == "__main__":
+    # Choose the `run` function based on what machine you're on.
+    # Change <username>
+
+    # Results will be at https://compy-dtn.pnl.gov/<username>/v2_4_0_all_sets/viewer/
```
forsyth2, line 228: `v2_5_0_all_sets`

examples/run_v2_5_0_all_sets_E3SM_machines.py
```diff
@@ -0,0 +1,236 @@
...
+    # run_compy('/compyfs/www/<username>/')
+
+    # Results will be at https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/<username>/v2_4_0_all_sets/viewer/
```
forsyth2, line 231: `v2_5_0_all_sets`

examples/run_v2_5_0_all_sets_E3SM_machines.py
```diff
@@ -0,0 +1,236 @@
...
+    # run_lcrc('/lcrc/group/e3sm/public_html/diagnostic_output/<username>/')
+
+    # Results will be at https://portal.nersc.gov/project/e3sm/<username>/v2_4_0_all_sets/viewer/
```
forsyth2, line 234: `v2_5_0_all_sets`

model_data_preprocess/postprocessing_E3SM_data_for_TC_analysis.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Get the list of files
+cd ${drc_in};eval ls ${drc_in}/${caseid}.$atm_name.h2.*{${start}..${end}}*.nc >${result_dir}inputfile_${file_name}.txt
+cd ${drc_in};eval ls ${caseid}.$atm_name.h2.*{${start}..${end}}*.nc >${result_dir}outputfile_${file_name}.txt
+#cd ${drc_in};eval ls ${drc_in}/${caseid}.cam.h2.*{${start}..${end}}*.nc >${result_dir}inputfile_${file_name}.txt
+#cd ${drc_in};eval ls ${caseid}.cam.h2.*{${start}..${end}}*.nc >${result_dir}outputfile_${file_name}.txt
+
+cd ${result_dir}
+# Detection threshold including:
+# The sea-level pressure (SLP) must be a local minimum; SLP must have a sufficient decrease (300 Pa) compared to surrounding nodes within 4 degree radius; The average of the 200 hPa and 500 hPa level temperature decreases by 0.6 K in all directions within a distance of 4Â° from the location o fSLP minima
```
forsyth2, line 65: `4Â°` -- is that correct or a formatting error/typo?

tests/complete_run.py
```diff
@@ -2,7 +2,7 @@
 import unittest
 
 # This test should be run with the latest E3SM Diags tutorial code.
-from examples.tutorials.tutorial_v2_4_0_all_sets_E3SM_machines import run_lcrc
+from examples.run_v2_5_0_all_sets_E3SM_machines import run_lcrc
```
forsyth2, line 5: As mentioned above, maybe we import a copy of the file that uses `e3sm_diags` instead of `acme_diags`?

forsyth2 (approved):
Looks good to me. I was able to run the example and generate the new diagnostics. I also reviewed the code. I'm not as familiar with the post-processing steps, however. The only issue I see is the `acme_diags` / `e3sm_diags` conflict.

examples/run_v2_5_0_all_sets_E3SM_machines.py
chengzhuzhang: yeah, there is some inconvenience during transition from importing `acme_diags` to `e3sm_diags`. The run script is fur v2_5_0, so i think it is fine to keep `import acme_diags`. I'm not sure what does it mean by "use a copy of this file that trades acme_diags for e3sm_diags"...

forsyth2: @chengzhuzhang I meant meant an exact copy of this file named something like `examples/run_v2_5_0_all_sets_E3SM_machines_updated_imports.py`, and then replace `acme_diags` calls with `e3sm_diags`. Then, in `tests/complete_run.py` use `from examples.run_v2_5_0_all_sets_E3SM_machines_updated_imports import run_lcrc`

I could do that in a follow-up PR, and update the test's expected images too.

model_data_preprocess/postprocessing_E3SM_data_for_TC_analysis.sh
chengzhuzhang: should replace with 4 degree radius.

examples/run_v2_5_0_all_sets_E3SM_machines.py
chengzhuzhang: Good reminder! the expected images need to be replaced as well. I think it makes sense to use this branch/PR to include the proposed change? Would you also help address the review comments? They are all good!

forsyth2: @chengzhuzhang yes I will push a commit addressing my suggestions.

forsyth2: @chengzhuzhang To update the tests, I need write permissions to `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets`

forsyth2:
@chengzhuzhang I just pushed a commit addressing my code review comments. I just labeled it "Address code review comments" since I assume you will merge this pull request with "Squash and merge". 

I do need write permissions to update `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets` however. As I mentioned in #506, I think we need to update the instructions to replace expected images to include a step on updating permissions. Something like `chmod -R g+wX /lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets`?

@tomvothecoder I had to use the `--no-verify` option to make the commit. Same Blocking IO error mentioned in #516.

chengzhuzhang:
@forsyth2 I made the folder writable by E3SM group. Not sure if other developers need to update permission after overwriting the folder. I think it depends. But I think the e3sm-diags developer team should take on the task to update this folder at least updating should be under our supervision...

forsyth2:
@chengzhuzhang 

>  I made the folder writable by E3SM group. 

I see that with `ls -l`, but `mv all_sets previous_output/all_sets_v250_20210608_1878984` still fails with `mv: cannot move 'all_sets' to 'previous_output/all_sets_v250_20210608_1878984/all_sets': Permission denied`

> Not sure if other developers need to update permission after overwriting the folder. 

Whoever is the last to update the expected images in `all_sets` would need to update the permissions to allow the next person to update them.

forsyth2:
> Permission denied

Never mind, `all_sets_v250_20210608_1878984` was apparently already in `previous_output`. I created a `all_sets_v250_20210608_1878984_2`.

chengzhuzhang:
cool, I saw that all_sets get moved as well. :)

forsyth2:
@chengzhuzhang Ok, I ran the `complete_run` test and it completed successfully.

What's strange however, is that the tropical cyclone analysis plots are generated but are not included in the viewer: Compare https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets/viewer/ with https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets/ (TC Analysis is listed in the latter but not the former)

Furthermore, the TC analysis plots did show up in my initial testing run (https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.forsyth2/v2_5_0_all_sets/viewer/). I don't think anything I changed in my commit would have removed the TC analysis plots from the viewer.

I also confirmed it's not just a browser cache problem by looking at my local file structure: `/lcrc/group/e3sm/public_html/diagnostic_output/ac.forsyth2/v2_5_0_all_sets/viewer/` (initial run's viewer) includes TC analysis, as does `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets` (latest run). However, TC analysis is NOT found in `/lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets/viewer` (latest run's viewer).

chengzhuzhang:
Hmm, since the only relevant change from your commit is to `import e3sm_diags` instead of `import acme_diags`, but i suppose you are actually testing v2_5_0? So I guess, if you could try use the original  run_v2_5_0 script with `import acme_diags` to see if that would work?

forsyth2:
>  if you could try use the original run_v2_5_0 script with import acme_diags to see if that would work?

Interesting; that worked. I'm unsure if the problem is actually using `e3sm_diags` rather than `acme_diags` though. I suspect that the actual problem is that I use the latest development environment to run with `e3sm_diags` but the latest Unified environment to run with `acme_diags`. Perhaps something that merged since the release of 2.5.0 erases the TC Analysis from the viewer?

chengzhuzhang:
I think let's use the original run_v2_5_0 with the `import acme_diags`. We can change in the next release.

forsyth2:
Well I don't think that will help much with the integration tests. The complete run is supposed to be testing the _latest_ version of the code -- that means using `e3sm_diags` rather than `acme_diags` and the latest development environment rather than the Unified environment. **I can make a separate issue** to investigate why TC analysis isn't showing up; that way we can merge this PR.

Of course, we'd still direct users to the not-updated v2_5_0 tutorial, since they do _not_ have the latest development environment.

chengzhuzhang:
I see, do you mean that with latest e3sm_diags development env, the TC won't show up? I'm a little confused about an early comment "I suspect that the actual problem is that I use the latest development environment to run with e3sm_diags but the latest Unified environment to run with acme_diags."

forsyth2:
> do you mean that with latest e3sm_diags development env, the TC won't show up?

Sorry, yes, correct:

## Case 1: `acme_diags`, Unified environment
1. Edit the import line of `tests/complete_run.py` to be `from examples.run_v2_5_0_all_sets_E3SM_machines import run_lcrc`
2. Run the following:
```
srun --pty --nodes=1 --time=01:00:00 /bin/bash
source /lcrc/soft/climate/e3sm-unified/load_latest_e3sm_unified_chrysalis.sh
python -m unittest tests/complete_run.py
```
3. Note that the test will fail because the images don't match up exactly. (For this case, I could have just updated the `main` method and run `python examples/run_v2_5_0_all_sets_E3SM_machines.py` directly. This is what a user running the v2.5.0 tutorial would do.)
4. The resulting viewer includes Annual Cycle Zonal Mean, ARM, and TC.

## Case 2: `e3sm_diags`, development environment
1. Edit the import line of `tests/complete_run.py` to be `from examples.run_v2_5_0_all_sets_E3SM_machines_updated_imports import run_lcrc`
2. Run the following:
```
srun --pty --nodes=1 --time=01:00:00 /bin/bash
conda activate e3sm_diags_env_dev_20210827 
pip install .
python -m unittest tests/complete_run.py
```
3. The tests will pass (i.e., the images outside of the viewer match up). The viewer includes Annual Cycle Zonal Mean and ARM, but _not_ TC.

## Conclusion
In writing this out I discovered something confusing: 4 out of 6 images are included in the TC images for Case 2 (the maps are missing), but all 6 are included in the TC images for Case 1. So something about Case 2 stops both the maps from showing up and the TC diags from showing up in the viewer at all.

In any case, I believe this problem is out of the scope of this pull request. My suggestion would be to merge this PR and I can look into this issue separately.

chengzhuzhang:
Thanks for the clarification. I agree we should merge this PR and fix the issue for TC viewer.

forsyth2:
I created #520 for the TC analysis issue.  

It looks like this branch is out-of-date with `master`. Would you like me to squash the commits, rebase, and merge?

chengzhuzhang:
Thanks for investigating the new issue. I will merge this branch once all checks pass.

PR #523, opened 2021-09-16
fix artificial warm bias zonal mean 2d

### LINKED ISSUE #522: Spurious lower tropospheric warm biases over Antarctica due to post-processing
It has been a persistent feature since v1that substantial lower-tropospheric warm biases are present in the zonal-pressure plots.
Below is [a difference plot of v2.LR.historical ensemble versus ERA-Interim](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/wlin/E3SMv2/v2.LR.HIST_ENS/e3sm_diags/180x360_aave/model_vs_obs_1985-2014/viewer/zonal_mean_2d/era-interim/t-global-era-interim/ann.html)
image

[Investigation led by @zhangshixuan1987 found it is mostly an artifact of post-processing](https://acme-climate.atlassian.net/wiki/spaces/ECG/pages/3048701961/Warm+bias+in+low+troposphere+over+southern+polar+region). 

The spurious biases are due to how the zonal mean difference plots are computed. Typically zonal means are computed first for model and obs (reanalysis), separately, which are then used to compute the differences. The root cause is the reanalysis data used utilize extrapolation to fill grids at all pressure levels, including below the ground surface, while the model data only include those above surface (pressure < PS). There would be no problem when at a certain level, the model has zero valid points (as seen in the blank space below around 875 hPa). When the model has grids partially masked out for a pressure level, the data points used to compute zonal mean could be very different between model and obs, let alone the reanalysis data contain extrapolated data.

The problem can be eliminated by computing the differences between model and obs first, using consistent masks, then deriving the zonal means.

chengzhuzhang:
Resolves #522 
Now, first take the difference between 3d data (to assure both datasets have same mask), then apply zonal average. 

Typically (oirinally) zonal means are computed first for model and obs (reanalysis), separately, which are then used to compute the differences.

The artificial warm bias is reduced after the fix.
Before:
Screen Shot 2021-09-16 at 2 10 47 PM
After:
Screen Shot 2021-09-16 at 2 11 09 PM

@wlin7 I think the result is align with @zhangshixuan1987's. Please let me know if it looks reasonable.

### DIFF
--- a/e3sm_diags/driver/zonal_mean_2d_driver.py
+++ b/e3sm_diags/driver/zonal_mean_2d_driver.py
@@ -156,6 +156,15 @@ def run_diag(parameter):
                 mv2_p = utils.general.convert_to_pressure_levels(
                     mv2, plevs, ref_data, var, season
                 )
+                # Regrid towards the lower resolution of the two
+                # variables for calculating the difference.
+                mv1_p_reg, mv2_p_reg = utils.general.regrid_to_lower_res(
+                    mv1_p,
+                    mv2_p,
+                    parameter.regrid_tool,
+                    parameter.regrid_method,
+                )
+                diff_p = mv1_p_reg - mv2_p_reg
 
                 mv1_p = cdutil.averager(mv1_p, axis="x")
                 mv2_p = cdutil.averager(mv2_p, axis="x")
@@ -187,7 +196,8 @@ def run_diag(parameter):
                     mv1_reg = mv1_p
                     mv2_reg = mv2_p
 
-                diff = mv1_reg - mv2_reg
+                # diff = mv1_reg - mv2_reg
+                diff = cdutil.averager(diff_p, axis="x")
                 metrics_dict = create_metrics(mv2_p, mv1_p, mv2_reg, mv1_reg, diff)
 
                 parameter.var_region = "global"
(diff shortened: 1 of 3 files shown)


e3sm_diags/driver/zonal_mean_2d_driver.py
```diff
@@ -187,7 +196,8 @@ def run_diag(parameter):
                     mv1_reg = mv1_p
                     mv2_reg = mv2_p
```
wlin7, line 197: Hi @chengzhuzhang , thanks for fixing this. The calculation procedure looks good and the new results are reasonable.

Don't know what is the purpose of computing mv1_p and mv2_p and the assignment to mv#_reg now. Are they still needed?

chengzhuzhang:
@wlin7 Good catch! Thank you for reviewing. mv#_reg are regridded data of mv1_p and mv2_p. They are needed to compute RMSE and CORR. I realized we need same masks for mv1_p and mv2_p before taking zonal mean as well to get the values of RMSE and CORR that are free of the artificial warm bias. I pushed a new commit to address this. The RMSE value dropped from 2.87 to 2.78 after same masks are applied to both model and obs data, see below. (Note that the new plots included part of stratosphere, therefore the metrics are slightly difference than plots here https://github.com/E3SM-Project/e3sm_diags/pull/523#issue-735826997) Let me know if it is okay to merge. 
Before:
Screen Shot 2021-09-21 at 1 41 38 PM


After
Screen Shot 2021-09-21 at 1 41 54 PM

wlin7 (approved):
Thanks for the further update and explanation, @chengzhuzhang . All look good to me. I think ok to merge.

PR #516, opened 2021-08-30
Add xml support for get_static_var

forsyth2:
Add multi-variable support for `xml`. Resolves #515.

PR #524, opened 2021-09-16
Use cartopy_offlinedata

### LINKED ISSUE #520: Tropical Cyclone images not showing up with development environment
As discussed in #514, Tropical Cyclone images do not show up with development environment. 4 out of 6 images show up in the `tc_analysis` subdirectory; the 2 maps do not. Furthermore, the TC diagnostics do not show up in the viewer. However, using the Unified environment, all 6 images are present and the TC diagnostics shows up in the viewer.

forsyth2:
Use `cartopy_offlinedata` to fix bug where Tropical Cyclone images don't show up. Resolves #520.

- [ ] After this merges, follow the steps at https://e3sm-project.github.io/e3sm_diags/_build/html/master/dev_guide/testing.html#complete-run-test to update the expected images for the `complete_run`test

### DIFF
--- a/conda/e3sm_diags_env.yml
+++ b/conda/e3sm_diags_env.yml
@@ -11,6 +11,7 @@ dependencies:
   - matplotlib=3.4.2
   - beautifulsoup4=4.9.3
   - cartopy=0.19.0
+  - cartopy_offlinedata
   - cdp=1.7.0
   - cdms2=3.1.5
   - cdtime=3.1.4
--- a/conda/e3sm_diags_env_dev.yml
+++ b/conda/e3sm_diags_env_dev.yml
@@ -11,6 +11,7 @@ dependencies:
   - matplotlib=3.4.2
   - beautifulsoup4=4.9.3
   - cartopy=0.19.0
+  - cartopy_offlinedata
   - cdp=1.7.0
   - cdms2=3.1.5
   - cdtime=3.1.4
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -31,6 +31,7 @@ requirements:
     - netcdf4
     - matplotlib
     - cartopy
+    - cartopy_offlinedata
     - beautifulsoup4
     - lxml
     - dask
--- a/examples/run_v2_5_0_all_sets_E3SM_machines_updated_imports.py
+++ b/examples/run_v2_5_0_all_sets_E3SM_machines_updated_imports.py
@@ -226,11 +226,10 @@ if __name__ == "__main__":
     # Change <username>
 
     # Results will be at https://compy-dtn.pnl.gov/<username>/v2_5_0_all_sets/viewer/
-    # run_compy('/compyfs/www/<username>/')
+    # run_compy("/compyfs/www/<username>/")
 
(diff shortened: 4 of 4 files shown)


conda/e3sm_diags_env.yml
```diff
@@ -11,6 +11,7 @@ dependencies:
   - matplotlib=3.4.2
   - beautifulsoup4=4.9.3
   - cartopy=0.19.0
+  - cartopy_offlinedata
```
forsyth2, line 14: @tomvothecoder Do we need to pin a specific version? It looks like v0.2.4 is the newest release on https://github.com/willirath/cartopy_offlinedata/releases

forsyth2 (commented):
@chengzhuzhang This is ready for review. It looks like we only need to load `cartopy_offlinedata` as part of the `conda` dependencies; no need to import it in every file using `cartopy`.

@tomvothecoder As with #514 and #516, I used `--no-verify` to make this commit.

chengzhuzhang (approved):

chengzhuzhang:
Thank you for the PR. I tagged @xylar for reviewing as well.

I haven't had to use `--no-verify` issue when committing. Not sure if it is LCRC specific ...

conda/e3sm_diags_env.yml
xylar: I would not pin a specific release.  The data doesn't typically change. The latest release was a small bug fix to how logging is handled when it warns you that your data is coming from this package rather than being downloaded.

xylar (approved):
Looks good to me!

tomvothecoder:
@forsyth2 @chengzhuzhang 

The `zppy` and `zstash` dev envs (both Python 3.7.10) run pre-commit fine on LCRC.

The error below that is outputted through `pre-commit` seems to be specific to LCRC servers and Python 3.9 in the`e3sm_dev` environment. I reinstalled the latest version of the `e3sm_diags` dev env on LCRC, but it outputs the same error. 

```bash
BlockingIOError: [Errno 11] Resource temporarily unavailable: '/gpfs/fs1/home/ac.tvo/.local/share/virtualenv/wheel/3.9/image/1/CopyPipInstall/pip-21.1.2-py3-none-any/pip/_vendor/distlib/w32.exe' -> '/gpfs/fs1/home/ac.tvo/.cache/pre-commit/repo7v9o5q3a/py_env-python3.9/lib/python3.9/site-packages/pip/_vendor/distlib/w32.exe'
```

Also, I tested the latest dev environment on my local machine and it works fine. 


What I suggest for now:
- Open an issue to further investigate this error and/or open up an issue directly on https://github.com/pre-commit/pre-commit
- Perform dev locally to `pre-commit` run, then pull those changes to test on `LCRC`, or use `no-verify` and fix the `pre-commit` errors that appear in the GH Actions build.

tomvothecoder:
Opened up an issue to address the LCRC + pre-commit issue here: https://github.com/E3SM-Project/e3sm_diags/issues/526

xylar:
@tomvothecoder, I believe I've seen issues like that frequently on Chrysalis.  It seems to have to do with recent versions of python and that file system.  We ended up having to use slow file copying as a fallback on Chrysalis in MPAS-Analysis because of this.

PR #525, opened 2021-09-16
refine plevs to include stratosphere

### LINKED ISSUE #492: Pressure-Latitude zonal mean contour plots don't cover stratosphere
The plots currently stop at 100 hPa, not covering the stratosphere. The stratospheric model results are important. In particular, the T in the southern polar region is useful to determine the ozone hole PSC T threshold. If the observational and difference plots cannot be extended, we at least need to plot the stratospheric model results.

chengzhuzhang:
Partly resolves #492 

By refining default setting for plevs. Stratosphere are being included in pressure-latitude plot. 

Before:
Screen Shot 2021-09-16 at 4 44 59 PM

After:

Screen Shot 2021-09-16 at 4 44 29 PM

### DIFF
--- a/e3sm_diags/parameter/zonal_mean_2d_parameter.py
+++ b/e3sm_diags/parameter/zonal_mean_2d_parameter.py
@@ -1,5 +1,3 @@
-import numpy
-
 from e3sm_diags.driver.utils.general import monotonic
 
 from .core_parameter import CoreParameter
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
Hi @tangq, 

This PR can partly cover what was requested in issue 492. But I couldn't find a way to have stratosphere showing in log, but troposphere in linear. We may need to generate new set for Stratosphere only. At least this PR is a good first step. Let me know if you'd like any changes...

tangq:
@chengzhuzhang , thanks for working on this. I'd suggest adding a new set for the stratosphere, but keep some overlapping areas, for example, trop plots: surface to 50 hPa; strat plots: 100 hPa to top. In this way, the transient is shown clearer.

For many variables, the stratospheric numbers are in a different scale than the tropospheric ones. Separating them will allow them emphasizing different aspects.

chengzhuzhang:
Yes, I'm thinking in the same line. We will define a separate set of contour level values for Stratospheric variables. I will just extend to 50 hPa for this PR. And then work on a new set for stratosphere starting only then.

chengzhuzhang:
@tangq  @golaz with a new commit, the pressure axis stops at 50hPa and shows top tick and ticklabel.
Screen Shot 2021-09-17 at 3 44 48 PM

chengzhuzhang:
Building error should be resolved after unit test image being updated (will do after fixing #523 )

PR #540, opened 2021-11-15
Optimize nco invocation

chengzhuzhang:
Following best practice of nco invocation.


czender (approved):
Looks good. As a reminder, this will work on EAM files with "eam" in the filename. It will break on EAM files with "cam" in the filename. To get the latter to work, you would need to specify `-P eam -m cam`.

chengzhuzhang:
Hey Charlie, thank you for reviewing. I'm a little confused about the note, from EAM files with 'cam' in the filename (e.g. most v1 data), would `-P cam` work? Could explain again the difference between `-P eam -m cam` and `-P cam`. Thanks!

czender:
Yes, if the filenames have "cam" in the name then `-P cam` will work, as will `-P eam -m cam`. I guess my point was that `-P eam` alone will not work on such files.

chengzhuzhang:
It's good to know. Thank you, Charlie.

PR #541, opened 2021-11-16
Document change from acme_diags

forsyth2:
Document the change from `acme_diags` to `e3sm_diags` in `v2.6.0`. Note: this change only affects the `master` docs, not the `v2.6.0` docs.


chengzhuzhang (commented):
Thanks for the PR, Ryan. How about something as follows:
Warning:

As of v2.6.0, e3sm_diags should be used as the module name instead of acme_diags. Instances of `acme_diags` in the Python import statements should be replaced accordingly.

forsyth2:
@chengzhuzhang Thanks, I'll change the wording to that and merge!

PR #544, opened 2021-12-08
Replace print statements with logger and execute integration tests explicitly over implicitly

tomvothecoder:
This PR adds a logger module that uses Python's logger to capture stderr outputs in a log file while maintaining them in the console. The log file is created in the root of where e3sm diags is installed, then transferred to the provenance directory at the end of the completed/failed diagnostic run. It also includes PR #546, which updates the execution of integration tests from implicit to explicit so that Python unittest treats the tests directory as a module to allow for absolute imports.

- Closes #148 

Example output in build under "Run Tests": https://github.com/E3SM-Project/e3sm_diags/runs/4537420648?check_suite_focus=true
Example log file in `/prov` directory: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.tvo/doc_examples/lat_lon_demo/prov/

Summary of Changes
- Add `logger.py` module with `custom_logger()` and `move_log_to_prov_dir()` functions
- Add try and except statement in `Run.run_diags()` to capture the traceback if e3sm diags crashes
  - Add call to `move_log_to_prov_dir()` in `Run.run_diags()`
- Replace all print statements with logger calls at appropriate logger levels
  - Why logger is preferred over print statements: https://stackoverflow.com/a/6918596
- Add function `run_cmd_and_pipe_stderr()` for running commands in integration tests and capturing stderr (for capturing the `results_dir` param)
  -  Before it was capturing stdout via print statements, but print statements have been deprecated in favor of using Python's logger.

Notes
- The log files don't capture print statements, so the `cdms2.mvCdms2Regrid` message below is not logged. This message should be either a `warning` or `logger.warning()` object instead.
https://github.com/CDAT/cdms/blob/cda99c21098ad01d88c27c6010d4affc8f621863/Lib/mvCdmsRegrid.py#L478-L482  
  ```
  WARNING: Edge bounds are the same. The results of conservative regridding will not conserve.
  coordMin =  -90.00, boundMin =  -90.00, coordMax =   90.00, boundMax =   90.00
  ```

PR #546, opened 2021-12-10
Update integration test execution and paths from implicit to explicit

### LINKED ISSUE #545: Update integration test paths from implicit to explicit
In PR #544, I discovered that you can't import modules that live in the `tests` dir in other `tests` modules because `test/test.sh` is calling the unit tests using `cd tests/integration`, rather than from the root of the project

https://github.com/E3SM-Project/e3sm_diags/blob/8852aefabd561c0ad27118be5aa91ab01f062321/tests/test.sh#L1-L19

For example in PR #544, I wrote a utility `tests.utils.run_command_and_save_stderr` and tried to import it into `tests/integration/test_all_sets.py`, but it will throw `ModuleNotFoundError`
```
from tests.utils import run_command_and_save_stderr
```
---
## Issues and their fixes:
1.  `tests/test.sh` performs `cd` calls into `tests/integration` and then executes the unit tests. This means the root directory is considered `tests/integration`, rather than the project root. As a result Python relies on implicit paths for imports (e.g., `from ..utils import run_command_save_stderr`). We need to change how unit tests are called to enable explicit imports.
    ```
    # Before (implicit imports only)
    cd integration && python -m unittest && cd ..
    
    # After (explicit imports allowed)
    python -m unittest discover -s "tests/integration"
    ```

2. Integration test data saves at whatever is considered the root folder. By changing the command above, the test data will be downloaded into the root of the project and not `tests/integration`. As a result we need to move the data after it finishes downloading to `tests/integration`

3. All paths referenced in integration tests and their configs need to be updated. For example, `"integration_tests_data"` and `"integration_tests_images"` to `"tests/integration/integration_tests_data"` and `"tests/integration/integration_test_images"`.

tomvothecoder:
This PR updates the commands in `test.sh` so that they are called explicitly, rather than using `cd tests` beforehand.

It also adds static string variables defining the paths and directories used in the integration test files to reduce repeated code.

- Closes #545 
- Extension of PR #544

### DIFF
--- a/tests/integration/all_sets.cfg
+++ b/tests/integration/all_sets.cfg
@@ -175,7 +175,6 @@ end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'TREFHT_201201_201312.nc'
 test_name = "system tests"
 variables = ["TREFHT"]
@@ -196,7 +195,6 @@ ref_end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'TREFHT_201201_201312.nc'
 test_name = "system tests"
 variables = ["TREFHT"]
@@ -214,7 +212,6 @@ end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'TREFHT_201201_201312.nc'
 test_name = "system tests"
 variables = ["TREFHT"]
@@ -233,7 +230,6 @@ end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'TREFHT_201201_201312.nc'
 test_name = "system tests"
 variables = ["TREFHT"]
@@ -253,7 +249,6 @@ ref_end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'TREFHT_201201_201312.nc'
 test_name = "system tests"
 variables = ["TREFHT"]
@@ -270,7 +265,6 @@ end_yr = '2013'
 test_data_path = 'tests/integration/integration_test_data'
 test_file = 'U_201301_201312.nc'
 reference_data_path = 'tests/integration/integration_test_data'
-
 ref_file = 'U_201301_201312.nc'
 test_name = "system tests"
 variables = ["U"]
--- a/tests/integration/all_sets.cfg
+++ b/tests/integration/all_sets.cfg
@@ -9,12 +9,12 @@ test_name = "system tests"
 short_test_name = "short_system tests"
 ref_name = "ERA-Interim"
 reference_name = "ERA-Interim Reanalysis 1979-2015"
-reference_data_path = "integration_test_data"
+reference_data_path = "tests/integration/integration_test_data"
 ref_file = "ta_ERA-Interim_ANN_198001_201401_climo.nc"
-test_data_path = "integration_test_data"
+test_data_path = "tests/integration/integration_test_data"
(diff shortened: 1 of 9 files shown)


tests/integration/test_dataset.py
```diff
@@ -6,20 +6,15 @@
 from e3sm_diags.derivations import acme as acme_derivations
 from e3sm_diags.driver.utils.dataset import Dataset
 from e3sm_diags.parameter.core_parameter import CoreParameter
-
-
-def get_abs_file_path(relative_path):
-    return os.path.join(os.path.dirname(os.path.abspath(__file__)), relative_path)
```
tomvothecoder, line 12: This function is no longer necessary since we replace it with an absolute path (`TEST_DATA_PATH`).

tests/integration/config.py
```diff
@@ -0,0 +1,8 @@
+# Paths and directories used in the integration test. Configurations in
+# `all_sets.cfg` and `all_sets_modified.cfg` should match these paths.
+TEST_ROOT_PATH = "tests/integration/"
+TEST_DATA_DIR = "integration_test_data"
+TEST_IMAGES_DIR = "integration_test_images"
+
+TEST_DATA_PATH = f"{TEST_ROOT_PATH}/{TEST_DATA_DIR}/"
+TEST_IMAGES_PATH = f"{TEST_ROOT_PATH}/{TEST_IMAGES_DIR}/"
```
tomvothecoder, line 8: This helps with reducing code redundancy throughout the integration test files.

There are probably effective ways to do this (e.g., using env vars), but it is more time consuming. If desired, it can be addressed in a future issue.

tomvothecoder:
I realized the CI/CD workflow isn't configured to run when the base branch of the PR isn't `master`. It should be okay though since the `pre-commit` and tests checks pass locally, and will run to the base branch in PR #544.

tomvothecoder:
Hey @chengzhuzhang and @forsyth2, this PR is ready for review whenever you have time. I decided to split this PR from PR #544 to make the review process clearer.

I also realized that the CI/CD workflow isn't configured to run when the base branch of the PR isn't `master.` It should be okay though since the pre-commit checks and unit tests pass locally. The build will run once this PR is merged to the base branch in PR #544.

tests/test.sh
```diff
@@ -2,18 +2,33 @@
...
+printf "==============================================\n"
+# Takes about four minutes
+python -m tests.integration.download_data
 
 # To use `scp`, run the following two lines instead. Be sure to change <username>. [Takes about two minutes]
 #scp -r <username>@blues.lcrc.anl.gov:/lcrc/group/e3sm/public_html/e3sm_diags_test_data/integration/integration_test_data integration_test_data
 #scp -r <username>@blues.lcrc.anl.gov:/lcrc/group/e3sm/public_html/e3sm_diags_test_data/integration/expected/integration_test_images integration_test_images
 
-# 4. Run integration tests
-cd integration && python -m unittest && cd ..
```
tomvothecoder, line 19: The reason why this worked before was because there was no command line argument being specified for running tests using a specific regex pattern (`test_*.py`). It was just `cd`ing into the `tests/integration` directory and running all tests in there.

tests/test.sh
```diff
@@ -2,18 +2,33 @@
...
+# Related lines of code : https://github.com/CDAT/cdp/blob/2917603097112f7db52be0bf7a2e766e14cc2e16/cdp/cdp_parser.py#L498
+command='python -m unittest tests/integration/'
+declare -a files=("test_all_sets" "test_dataset" "test_diags" "test_run")
+for file in "${files[@]}"
+    do
+        file="${file}.py"
+        printf "\n${file}\n"
+        printf "~~~~~~~~~~~~~~~~~~~~\n"
+        eval "${command}${file}"
+done
```
tomvothecoder, line 34: I'm not fluent in bash, this was a result of Google and Stack Overflow.

tests/integration/all_sets.cfg
```diff
@@ -221,14 +225,15 @@ sets = ["enso_diags"]
...
 seasons = ["ANN"]
 test_start_yr = '2012'
 ref_start_yr = '2012'
 end_yr = '2013'
-test_data_path = 'integration_test_data'
+test_data_path = 'tests/integration/integration_test_data'
 test_file = 'TREFHT_201201_201312.nc'
-reference_data_path = 'integration_test_data'
+reference_data_path = 'tests/integration/integration_test_data'
+
```
forsyth2, line 236: Are these extra blank lines from the formatter? If not, I think they should be removed.

forsyth2 (approved):
Looks good, thanks! I also tested on Chrysalis.

tests/integration/all_sets.cfg
tomvothecoder: Thanks for the catch, it was most likely from an extra whitespace in my find and replace haha. 
Whitespaces have been reverted.

forsyth2: Thanks!

Issue #549, opened 2022-01-10
Improvements to viewer for e3sm_diags for cmip


tomvothecoder:
Per the 1/7/22 meeting:
- Add banner with legend for realization column (and others if needed)
- Make table column headers sticky

This issue involves looking into the viewer code and seeing what CSS framework is used (if any), and how the HTML is structured. If it is simple, might be able to add a CSS attribute to make the table column headers stick as the user scrolls.

chengzhuzhang:
Hey Tom, thanks for creating this issue. The viewer to show all CMIP results was generated by a separate script Chris developed. It is currently not included in e3sm_diags code base. I will include those scripts so that you could help enhance.

chengzhuzhang:
Hey @tomvothecoder, the steps to generate the viewer is as follows (note you would need `jinja2` in your Python env):
```
cd /var/www/acme/acme-diags/e3sm_diags_for_cmip
python generate_page.py
```
Let me know if there is permission issue with the data. Thanks!

tomvothecoder:
Closing this issue since we're moving away from CDP Output Viewer (#628).

PR #555, opened 2022-01-20
Add e3sm_diags for CMIP example workflow

chengzhuzhang:
This PR is to add example workflow to run E3SM Diags looping over CMIP data archived at LLNL.


chengzhuzhang:
Hey @tomvothecoder, as we discussing in #549 . The generate_html.py scirpt could be enhanced. Since there is no source code change, i will just go ahead and merge this PR. I will create instruction to reproduce the viewer on ACME1 for you to take over.

Issue #560, opened 2022-02-07
Regridding artifacts in model-vs-model comparison


oksanaguba:
The issue came up when i made model-vs-model climo for a run with ne30pg2 grid (new run, made with a mapping file below) against a run with ne30np4 grid (baseline, made with a different mapping file, also below). See blue contours 
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/onguba/theta/theta-cpterms-jan27.144547.0002-0006/def/viewer/lat_lon/model_vs_model/t-850mb-global/ann.html

When i made climo against a ne30pg2 baseline run, the contours are gone:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/onguba/theta/theta-cpterms-jan27.144547-against-pg.0002-0006/def/viewer/lat_lon/model_vs_model/t-850mb-global/ann.html

Maps that i used for both grids are
map_ne30np4_to_fv128x256_aave.20160301.nc  
map_ne30pg2_to_cmip6_180x360_aave.20200201.nc

Since maps regular grids differ, may it be expected behavior?

wlin7:
Hi @oksanaguba ,try to understand the issue.  Are the two baseline runs themselves actually different -- the middle panels of the two pages, one with ne30np4, the other with ne30pg2?

oksanaguba:
yes, baselines differ, np4 and pg2, the run to investigate is pg2, thanks.

wlin7:
Thanks. In that case, we would expect the results to be different; and the differences between the two difference plots (3rd panel in each) are due to differences in the baseline simulations. Therefore, the differences are a result of model configuration: the treatment of physics columns and the cumulative interaction between physics and dynamics over 6 years of simulations. Regridding (with e3sm_diag) itself is fine, with proper mapping files for the corresponding output data. But please further elaborate if I misunderstood the issue.

oksanaguba:
Thanks, Wuyin, i agree it is not such a problem if one can use proper baselines. However, in this case it seems to me that blue contours appear close to places where one of files does not have data (which are marked in white). So, extrema are created for regions where one of files does not have data. If this is the case, this seems to be wrong behavior?  This example is also problematic because the destination grids do not match. 

In other words, to me it seems that blue contours are due to nonexistent data in the first two plots, not because there is difference between 2 simulations.

wlin7:
[...]

Issue #411, opened 2021-03-19
Errors around coast lines on SST plots with v1-HR model output


golaz:
I ran e3sm_diags on output from v1-HR, regridded using `map_ne120np4_to_cmip6_720x1440_nco.20190601.nc`.

The SST plots show erroneous values near the coastlines, leading to a very large RMSE error:

![image](https://user-images.githubusercontent.com/15743046/111833190-118c6500-88af-11eb-99f2-ec6039d09655.png)

Full E3SM Diags output: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.golaz/E3SM/v1-HR/20210112.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG.unc06/e3sm_diags/720x1440_nco/model_vs_obs_0137-0155/viewer/

Here is my configuration to reproduce on chrysalis using zppy:

```
[default]
input = /lcrc/group/e3sm/ac.ndkeen/scratch/chrys/E3SM_simulations/production-unc06
input_subdir = run
output = /lcrc/group/e3sm/ac.golaz/E3SM_analysis/20210112.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG.unc06
case = 20210112.A_WCYCL1950S_CMIP6_HR.ne120_oRRS18v3_ICG.unc06
www = /lcrc/group/e3sm/public_html/diagnostic_output/ac.golaz/E3SM/v1-HR
partition = compute
e3sm_unified = latest

[climo]
active = True
years = "137-155",

  [[ atm_monthly_720x1440_nco ]]
  input_files = "cam.h0"
  mapping_file = /home/ac.zender/data/maps/map_ne120np4_to_cmip6_720x1440_nco.20190601.nc

[ts]
active = True
years = "137:155:1",

  [[ atm_monthly_720x1440_nco ]]
  input_files = "cam.h0"
  mapping_file = /home/ac.zender/data/maps/map_ne120np4_to_cmip6_720x1440_nco.20190601.nc
  vars = "FSNTOA,FLUT,FSNT,FLNT,FSNS,FLNS,SHFLX,QFLX,TAUX,TAUY,PRECC,PRECL,PRECSC,PRECSL,TS,TREFHT,CLDTOT,CLDHGH,CLDMED,CLDLOW"

  [[ atm_monthly_glb ]]
  input_files = "cam.h0"
  mapping_file = "glb"
[...]

Issue #175, opened 2018-07-30
Inconsistent masks for areas for 850mb plot


chengzhuzhang:
Reported by @PeterCaldwell 
"the bias calculation for 850mb seems to use inconsistent masks for areas where 850mb is below the land surface, causing rings of massive bias values. See:
 
http://portal.nersc.gov/project/acme/coupled/tuning_tests/20180410.A_WCYCL1950_HR.ne120_oRRS18v3_ICG.theta/e3sm_diags_yr6-10/viewer/latitude-longitude/merra2/t-850mb-global-merra2/ann.html
"

"I’m using this file to regrid: map_ne120np4_to_181x360_aave.nc"

chengzhuzhang:
After reviewing some test data, we found that the inconsistent mask only occur when conservative method is applied when to regrid higher res data (obs data in this case) to low res data (model data) before comparing. 
Results using **conservative** method, 
https://acme-viewer.llnl.gov/shaheen2/all_sets_cdat8_esmf_7.1.0r-TRY2/lat_lon/MERRA2/MERRA2-T-850-ANN-global.png
Results using **bilinear** regridding method,
https://acme-viewer.llnl.gov/shaheen2/all_sets_cdat8_esmf_7.1.0r-linear/lat_lon/MERRA2/MERRA2-T-850-ANN-global.png

The treatment of masking in conservative method should plays a role in this issue. @dnadeau4 would you maybe comment on this issue? Thanks!

dnadeau4:
@taylor13 and I are working on a solution for conservative masking.  The one I implemented from NCAR does not satisfied your need.   (BTW I cannot access these pictures...)

chengzhuzhang:
Thank you denis! here is the access info to the pics..
Username: acme
Password: AwesomeModel

PR #574, opened 2022-03-21
Set regrid method to bilinear for a subset of variables with masks or missing data

chengzhuzhang:
This PR is to temporarily address a long-standing issue reported in #175  #411 #560  , where errors around edges of land/sea mask or missing data. The root cause is the treatment of masking in conservative method of cdms2.
For now, I have set variables with masks/missing data to use billinear regridding. TREFHT with ocean mask, TAUXY with land mask, T at 850mb, etc. 
To run with a sample simulation, the errors are gone. 
[conserv method](https://portal.nersc.gov/project/e3sm/zhang40/tests/e3sm_diags_regrid_consv/viewer/) vs [bilinear method](https://portal.nersc.gov/project/e3sm/zhang40/tests/e3sm_diags_regrid_bilinear/viewer/) comparing impacted variables.  
here gives ANN metrics difference, as a reference: https://www.diffchecker.com/iNS5B5gz 

We will replace the cdms2 regridding with xcdat/xesmf tool in the coming a few months.

### DIFF
--- a/e3sm_diags/driver/default_diags/lat_lon_model_vs_model.cfg
+++ b/e3sm_diags/driver/default_diags/lat_lon_model_vs_model.cfg
@@ -18,6 +18,7 @@ variables = ["SST"]
 seasons = ["ANN", "DJF", "MAM", "JJA", "SON"]
 contour_levels = [-1, 0, 1, 3, 6, 9, 12, 15, 18, 20, 22, 24, 26, 28, 29]
 diff_levels = [-2.5, -2, -1.5, -1, -0.5, -0.1, 0.1, 0.5, 1, 1.5, 2.0, 2.5]
+regrid_method = "bilinear"
 
 
 [#]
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
Hey @golaz I should have included all impacted variables to use bilinear instead. Let me know if this PR is okay to merge now. Thank you!

golaz (approved):
@chengzhuzhang : thanks for implementing this. It will be very useful to have ERA5 comparison and fix the coastline issues in some fields.

chengzhuzhang:
Thank you, Chris! @golaz 

Just a note here, we document a complete run with the same example datasets when large code changes occur.
For instance, before merging this PR, the complete run (including ERA variables) is documented [here](https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/unit_test_complete_run/expected/previous_output/all_sets_v2_6_1_20220328_d62f554/viewer/)
With this PR, the result is [here](https://web.lcrc.anl.gov/public/e3sm/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets/viewer/) (this link will point to updated run though when new results-changing run is made.)

Issue #563, opened 2022-02-22
Add metrics checker to integration tests


forsyth2:
Add metrics checker to integration tests

forsyth2:
Such a test would have caught #561. Equivalent issue for `zppy`: https://github.com/E3SM-Project/zppy/issues/199

PR #575, opened 2022-03-30
Metrics checker

chengzhuzhang:
Add a metrics checker to find lines of metrics being changed between two e3sm diags lat-lon runs.


chengzhuzhang:
@forsyth2 Hey Ryan, I think this small utility is handy for a quick check. And it potentially can help address #563

auxiliary_tools/metrics_checker.py
```diff
@@ -0,0 +1,60 @@
+#!usr/bin/env python
+from __future__ import division, print_function
```
forsyth2, line 2: Aren't we using `python3` now? I don't think we should still have `__future__` imports then.

auxiliary_tools/metrics_checker.py
```diff
@@ -0,0 +1,60 @@
...
+
+import argparse
+import os
+
+"""
+Usage: metrics_checker.py [options]
+Example: python metrics_checker.py -t /lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/previous_output/all_sets_v2_6_1_20220328_d62f554/ -r /lcrc/group/e3sm/public_html/e3sm_diags_test_data/unit_test_complete_run/expected/all_sets/
```
forsyth2, line 9: This example works (albeit with a few `Failed to open file` errors). 

However, I ran the tests and then `python metrics_checker.py -t ../tests/integration/all_sets_results_test/ -r ../tests/integration/integration_test_images/` which isn't that helpful because `viewer` doesn't have a `table-data` subdirectory. What needs to happen for that subdirectory to exist?

auxiliary_tools/metrics_checker.py
```diff
@@ -0,0 +1,60 @@
...
+def compare_metrics(ref_path, test_path, season):
+    fref = os.path.join(ref_path, "viewer/table-data", f"{season}_metrics_table.csv")
+    ftest = os.path.join(test_path, "viewer/table-data", f"{season}_metrics_table.csv")
+    try:
+        with open(fref, "r") as ref, open(ftest, "r") as test:
+            file_ref = ref.readlines()
+            file_test = test.readlines()
+            for line in file_ref:
+                if line not in file_test:
+                    print(f"Found difference in {season}", line)
```
forsyth2, line 53: It might be useful to also check the reverse -- if a line in `file_test` isn't in `file_ref`.

forsyth2 (approved):

tomvothecoder:
@chengzhuzhang I rebased this branch on the latest `master` with PR #576 to fix `black`, and pushed the `pre-commit` fixes.

auxiliary_tools/metrics_checker.py
chengzhuzhang: I guess this is expected, the results under both path here mentioned are not a typical e3sm_diags run directory.

auxiliary_tools/metrics_checker.py
chengzhuzhang: This was intentional. If there are additional lines file_test, then there was no reference line to compare with...but it is a good ideal to print the additional variables, I think.

forsyth2: Ah ok, that makes sense, but yes it would be nice to know *how* the metrics don't match up rather than just saying a difference exists.

chengzhuzhang:
@forsyth2 Hey Ryan, I have included changes suggested by your review. Since you have approved the PR, so i just went ahead and merge. Let me know if you get a chance to use this tool and we can further improve it.

Issue #536, opened 2021-10-22
Enable plotting when only test/model data are available


chengzhuzhang:
This feature will allow visualization of fields when observation data are not currently available.

chengzhuzhang:
This feature is useful land variables and aerosol variables when available obs are limited.

chengzhuzhang:
I'm trying to find a less intrusive way (minimal code changes) to implement this for lat-lon set (and potentially for other sets as well). One possibility is to modify the lat-lon driver, so that when the reference dataset (mv2 as referred by following codes) is not available, make mv2 a copy of mv1, and add an attribute to label mv2 as a duplicate of mv1. The label can be used to disable visualizing or computing metrics relevant to mv2.  

https://github.com/E3SM-Project/e3sm_diags/blob/a776ece46f9b0cb2154011fb39373445610a4590/e3sm_diags/driver/lat_lon_driver.py#L89-L90

@tomvothecoder @forsyth2 or @golaz please let me know if it sounds okay, or propose a better solution...

chengzhuzhang:
resolved by #584

PR #584, opened 2022-04-27
Support Model only analysis

chengzhuzhang:
This PR is to address #536 
Enable plotting and generating metrics tables when reference data is not present. 
For instances:
1. Following single command line, would plot the precipitation lat-lon plot without including a reference data:
e3sm_diags lat_lon --no_viewer --sets 'lat_lon' --variables 'PRECT' --seasons 'ANN'  --test_name '20160520.A_WCYCL1850.ne30_oEC.edison.alpha6_01' --test_colormap 'WhiteBlueGreenYellowRed.rgb' --test_data_path '/Users/zhang40/Documents/ACME_simulations/' --reference_data_path '/Users/zhang40/Documents/ACME_simulations/obs_for_e3sm_diags/climatology' --results_dir '/Users/zhang40/Downloads/lat_lon_model_vs_none' --output_format_subplot 'pdf'

2. Users can use following .cfg file to create diags with a combination of variables, with or without an observation dataset.
[example output](https://portal.nersc.gov/cfs/e3sm/zhang40/tests/lat_lon_test_model_only/viewer/)
```
[Diags]
sets = ['lat_lon']
case_id = "model_only"
variables = ["AODVIS"]
seasons = ["ANN"]
regions = ["global"]

[Diags2]
sets = ['lat_lon']
case_id = "model_only"
variables = ["PRECL"]
seasons = ["ANN"]
regions = ["global"]

[Diags3]
sets = ['lat_lon']
case_id = "GPCP_v2.3"
ref_name = "GPCP_v2.3"
variables = ["PRECT"]
seasons = ["ANN"]
regions = ["global"]

### DIFF
--- a/e3sm_diags/driver/lat_lon_driver.py
+++ b/e3sm_diags/driver/lat_lon_driver.py
@@ -8,41 +8,92 @@ import cdms2
 import e3sm_diags
 from e3sm_diags.driver import utils
 from e3sm_diags.logger import custom_logger
-from e3sm_diags.metrics import corr, max_cdms, mean, min_cdms, rmse, std
+from e3sm_diags.metrics import corr, mean, rmse, std
 from e3sm_diags.plot import plot
 
 logger = custom_logger(__name__)
 
 
+def create_and_save_data_and_metrics(parameter, mv1_domain, mv2_domain):
+    if parameter.ref_name != "":
+        # Regrid towards the lower resolution of the two
+        # variables for calculating the difference.
+        mv1_reg, mv2_reg = utils.general.regrid_to_lower_res(
+            mv1_domain,
+            mv2_domain,
+            parameter.regrid_tool,
+            parameter.regrid_method,
+        )
+
+        diff = mv1_reg - mv2_reg
+    else:
+        mv2_domain = None
+        mv2_reg = None
+        mv1_reg = mv1_domain
+        diff = None
+
+    metrics_dict = create_metrics(mv2_domain, mv1_domain, mv2_reg, mv1_reg, diff)
+
+    # Saving the metrics as a json.
+    metrics_dict["unit"] = mv1_domain.units
+
+    fnm = os.path.join(
+        utils.general.get_output_dir(parameter.current_set, parameter),
+        parameter.output_file + ".json",
+    )
+    with open(fnm, "w") as outfile:
+        json.dump(metrics_dict, outfile)
+
+    logger.info(f"Metrics saved in {fnm}")
+
+    plot(
+        parameter.current_set,
+        mv2_domain,
+        mv1_domain,
+        diff,
+        metrics_dict,
+        parameter,
+    )
+    utils.general.save_ncfiles(
+        parameter.current_set,
+        mv1_domain,
+        mv2_domain,
+        diff,
+        parameter,
+    )
+
+
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     """Creates the mean, max, min, rmse, corr in a dictionary"""
+    # For input None, metrics are instantiated to 999.999.
+    # Apply float() to make sure the elements in metrics_dict are Jason serializable, i.e. np.float64 type is JSON serializable, but not np.float32.
     missing_value = 999.999
     metrics_dict = {}
     metrics_dict["ref"] = {
-        "min": float(min_cdms(ref)) if ref is not None else missing_value,
-        "max": float(max_cdms(ref)) if ref is not None else missing_value,
+        "min": float(ref.min()) if ref is not None else missing_value,
+        "max": float(ref.max()) if ref is not None else missing_value,
         "mean": float(mean(ref)) if ref is not None else missing_value,
     }
     metrics_dict["ref_regrid"] = {
-        "min": float(min_cdms(ref_regrid)) if ref_regrid is not None else missing_value,
-        "max": float(max_cdms(ref_regrid)) if ref_regrid is not None else missing_value,
+        "min": float(ref_regrid.min()) if ref_regrid is not None else missing_value,
+        "max": float(ref_regrid.max()) if ref_regrid is not None else missing_value,
         "mean": float(mean(ref_regrid)) if ref_regrid is not None else missing_value,
         "std": float(std(ref_regrid)) if ref_regrid is not None else missing_value,
     }
     metrics_dict["test"] = {
-        "min": float(min_cdms(test)),
-        "max": float(max_cdms(test)),
+        "min": float(test.min()),
+        "max": float(test.max()),
         "mean": float(mean(test)),
     }
     metrics_dict["test_regrid"] = {
-        "min": float(min_cdms(test_regrid)),
-        "max": float(max_cdms(test_regrid)),
+        "min": float(test_regrid.min()),
+        "max": float(test_regrid.max()),
         "mean": float(mean(test_regrid)),
         "std": float(std(test_regrid)),
     }
     metrics_dict["diff"] = {
-        "min": float(min_cdms(diff)) if diff is not None else missing_value,
-        "max": float(max_cdms(diff)) if diff is not None else missing_value,
+        "min": float(diff.min()) if diff is not None else missing_value,
+        "max": float(diff.max()) if diff is not None else missing_value,
         "mean": float(mean(diff)) if diff is not None else missing_value,
     }
     metrics_dict["misc"] = {
@@ -93,7 +144,7 @@ def run_diag(parameter):  # noqa: C901
             mv1 = test_data.get_climo_variable(var, season)
             try:
                 mv2 = ref_data.get_climo_variable(var, season)
-            except Exception:
+            except (RuntimeError, IOError):
                 mv2 = mv1
                 logger.info("Can not process reference data, analyse test data only")
                 parameter.ref_name = ""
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: e3sm_diags/driver/lat_lon_driver.py [hunk 1]
```diff
@@ -87,45 +91,20 @@ def run_diag(parameter):
             parameter.var_id = var
 
             mv1 = test_data.get_climo_variable(var, season)
-            mv2 = ref_data.get_climo_variable(var, season)
+            try:
+                mv2 = ref_data.get_climo_variable(var, season)
+            except Exception:
+                mv2 = mv1
```
chengzhuzhang, line 97: using "mv2=mv1" is a convenient way to have a fake datasets with all mv1 attributes and values..

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -17,17 +16,18 @@
 
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
```
chengzhuzhang, line 17: When an input is None, asign missing values to the metrics.

chengzhuzhang:
@tomvothecoder let me know if you have any questions for a review. Thanks!

e3sm_diags/driver/lat_lon_driver.py
xylar: Just a note to say I've learned the hard way that a general `Exception` can be risky. It catches a `SyntaxError` in the code itself.

tomvothecoder, line 97:
```suggestion
            try:
                mv2 = ref_data.get_climo_variable(var, season)
            except RuntimeError:
                mv2 = mv1
```

As Xylar noted, it is best practice to be as specific as possible with Python exceptions (https://docs.python.org/3/library/exceptions.html).

The `.get_climo_variable()` function raises  `RuntimeError`  in multiple conditionals:

https://github.com/E3SM-Project/e3sm_diags/blob/b15a05feb08539e9c71ef0628720a90a415c4dc0/e3sm_diags/driver/utils/dataset.py#L128-L131

https://github.com/E3SM-Project/e3sm_diags/blob/b15a05feb08539e9c71ef0628720a90a415c4dc0/e3sm_diags/driver/utils/dataset.py#L162-L166

_Side-note: The exceptions in `.get_climo_variable()` should be `TypeError` instead of `RuntimeError`._

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -17,17 +16,18 @@
...
     """Creates the mean, max, min, rmse, corr in a dictionary"""
+    missing_value = 999.999
```
tomvothecoder, line 19: It is a good idea to add a comment explaining why `missing_value` is instantiated to 999.999.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -177,22 +156,30 @@ def run_diag(parameter):
...
 
-                        # Regrid towards the lower resolution of the two
-                        # variables for calculating the difference.
-                        mv1_reg, mv2_reg = utils.general.regrid_to_lower_res(
-                            mv1_domain,
-                            mv2_domain,
-                            parameter.regrid_tool,
-                            parameter.regrid_method,
-                        )
+                        if parameter.ref_name != "":
```
tomvothecoder, line 159: The logic in lines 159-198 and 231-266 are the same, so I would extract them as an internal function to avoid code duplication and allow for re-usability.

Example: 

```
def _create_and_save_metrics(parameter, mv1, mv1_domain, mv2_domain):
    if parameter.ref_name != "":
        # Regrid towards the lower resolution of the two
        # variables for calculating the difference.
        mv1_reg, mv2_reg = utils.general.regrid_to_lower_res(
                            mv1_domain,
                            mv2_domain,
                            parameter.regrid_tool,
                            parameter.regrid_method,
                        )

        diff = mv1_reg - mv2_reg
        metrics_dict = create_metrics(
                            mv2_domain, mv1_domain, mv2_reg, mv1_reg, diff
                        )
    else:
        mv2_domain = None
        mv2_reg = None
        diff = None
        metrics_dict = create_metrics(
                            mv2_domain, mv1, mv2_reg, mv1, diff
                        )

    # Saving the metrics as a json.
    metrics_dict["unit"] = mv1.units
    fnm = os.path.join(
                        utils.general.get_output_dir(parameter.current_set, parameter),
                        parameter.output_file + ".json",
                    )
    with open(fnm, "w") as outfile:
        json.dump(metrics_dict, outfile)

    # Get the filename that the user has passed in and display that.
    fnm = os.path.join(
                        utils.general.get_output_dir(parameter.current_set, parameter),
                        parameter.output_file + ".json",
                    )
    logger.info(f"Metrics saved in {fnm}")
    return mv2_domain, diff, metrics_dict
```

Then replace lines 159-198 and 231-254 with:
```
                    mv2_domain, diff, metrics_dict = _create_and_save_metrics(
                        parameter, mv1, mv1_domain, mv2_domain
                    )
```

There are better ways of writing the new `_create_and_save_metrics()` function (e.g., break that up into smaller functions), but I'll leave that up to you. It would make unit testing easier in the future if tests aren't written in this PR.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -208,7 +195,7 @@ def run_diag(parameter):
                             ),
                             parameter.output_file + ".json",
                         )
-                        print(f"Metrics saved in: {fnm}")
+                        logger.info(f"Metrics saved in: {fnm}")
```
tomvothecoder, line 199: I would move lines 200-215 outside of the conditional (line 289) and delete lines 268-283 since the logic is the same.

tomvothecoder (changes requested):
Hi Jill, I had some suggestions that you might want to consider.

Let me know what you think. I'm also happy to see `MV2` is removed here!

chengzhuzhang:
@xylar @tomvothecoder I just get back to work on this PR. Thank you both for reviewing, I will address the comments soon.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -4,55 +4,110 @@
 import os
 
 import cdms2
-import MV2
 
 import e3sm_diags
 from e3sm_diags.driver import utils
 from e3sm_diags.logger import custom_logger
-from e3sm_diags.metrics import corr, max_cdms, mean, min_cdms, rmse, std
+from e3sm_diags.metrics import corr, mean, rmse, std
```
chengzhuzhang, line 11: max_cdms and min_cdms no longer need to be imported. They are still being used in other drivers, we can consider remove all instances in a future house cleaning PR.

e3sm_diags/driver/lat_lon_driver.py
chengzhuzhang: The missing value 999.999 is used in another application of e3sm_diags for cmip model comparison by Chris, so I just went ahead and adopted it. There might be other alternatives, i.e. use NAN, but that will require more code change downstream.

chengzhuzhang (commented):
Hey @tomvothecoder I should have addressed most comments. I found that float() is applied to metrics numbers to ensure that elements of the dictionary `metrics_dict` is json serializable, so I kept the float() instances and left a comment on why float() is kept.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -4,55 +4,110 @@
...
 from e3sm_diags.plot import plot
 
 logger = custom_logger(__name__)
 
 
+def create_and_save_data_and_metrics(parameter, mv1_domain, mv2_domain):
```
chengzhuzhang, line 17: I didn't make it an internal function, since it could be used by other drivers when we get to refactor those, also lat_lon_land uses it as a base function.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -4,55 +4,110 @@
...
+        mv2_domain,
+        diff,
+        parameter,
+    )
+
+
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     """Creates the mean, max, min, rmse, corr in a dictionary"""
+    # For input None, metrics are instantiated to 999.999.
+    # Apply float() to make sure the elements in metrics_dict are Jason serializable, i.e. np.float64 type is JSON serializable, but not np.float32.
```
chengzhuzhang, line 69: Leave a comment on why float() is kept.

e3sm_diags/driver/lat_lon_driver.py
chengzhuzhang: Now use exception (RuntimeError, IOError) instead.

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder, line 69:
```suggestion
    # Apply float() to make sure the elements in metrics_dict are JSON serializable, i.e. np.float64 type is JSON serializable, but not np.float32.
```

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: Private (internal) functions usually refer to functions that are only called internally by other functions. If it is also intended to be available for the end-user, then it is public.

I don't think it is a big deal in this case though so no need to change it. We can update it later if you want.

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: I think you mean JSON and not Jason haha. I committed the change for you.

tomvothecoder (approved):

e3sm_diags/driver/lat_lon_driver.py
chengzhuzhang: Thanks for catching this...

PR #585, opened 2022-04-28
Add lat_lon_land

### LINKED ISSUE #566: Add model versus model example from ELM runs
Following diags runs provided by @dmricciuto, ex.:https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.ricciuto/GSWP3comp/lat_lon/viewer/

chengzhuzhang:
This PR adds a lat_lon set for land variables, this will be one step towards resolving #566 
[example output](https://portal.nersc.gov/cfs/e3sm/zhang40/tests/lat_lon_test_land/viewer/)

### DIFF
--- a/e3sm_diags/parameter/__init__.py
+++ b/e3sm_diags/parameter/__init__.py
@@ -27,4 +27,5 @@ SET_TO_PARAMETERS = {
     "arm_diags": ARMDiagsParameter,
     "tc_analysis": TCAnalysisParameter,
     "annual_cycle_zonal_mean": ACzonalmeanParameter,
+    "lat_lon_land": CoreParameter,
 }
--- a/e3sm_diags/parameter/__init__.py
+++ b/e3sm_diags/parameter/__init__.py
@@ -4,6 +4,7 @@ from .arm_diags_parameter import ARMDiagsParameter
 from .core_parameter import CoreParameter
 from .diurnal_cycle_parameter import DiurnalCycleParameter
 from .enso_diags_parameter import EnsoDiagsParameter
+from .lat_lon_land_parameter import LatLonLandParameter
 from .meridional_mean_2d_parameter import MeridionalMean2dParameter
 from .qbo_parameter import QboParameter
 from .streamflow_parameter import StreamflowParameter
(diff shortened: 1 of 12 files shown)


e3sm_diags/parameter/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .core_parameter import CoreParameter
 from .diurnal_cycle_parameter import DiurnalCycleParameter
 from .enso_diags_parameter import EnsoDiagsParameter
+from .lat_lon_land_parameter import LatLonLandParameter
```
chengzhuzhang, line 7: In order to co-operate both lat-lon and lat-lon-land, I tried to use the CoreParameter for lat_lon_land (with different test path and test name), but it didn't work because there was a parameter check which doesn't allow same type of parameters (https://github.com/E3SM-Project/e3sm_diags/blob/a776ece46f9b0cb2154011fb39373445610a4590/e3sm_diags/run.py#L48-L53), not sure why this needs to be restricted, but I went ahead to added a new parameter object.

tomvothecoder: My initial impression is that you can't have "two or more parameters of the same type" because those parameters (aka class attributes) are copied from the parent class (`CoreParameter`) to the child classes (such as `LatLonLandParameter`). A class attribute can't be defined twice, just like a variable.

`self._add_parent_attrs_to_children()` does the copying

https://github.com/E3SM-Project/e3sm_diags/blob/a776ece46f9b0cb2154011fb39373445610a4590/e3sm_diags/run.py#L48-L55

https://github.com/E3SM-Project/e3sm_diags/blob/a776ece46f9b0cb2154011fb39373445610a4590/e3sm_diags/run.py#L94-L104

tomvothecoder (approved):
Looks good to me

e3sm_diags/parameter/__init__.py
chengzhuzhang: Okay, thanks, it seems there has to be a new parameter object for lat-lon-land in this framework. I will go ahead and merge.

Issue #597, opened 2022-06-12
Error in complete lat_lon run on Chrysalis suggesting `model_only` feature coflicts with `cmip6_comparison` viewer


chengzhuzhang:
Testing 2.7.0rc1 within e3sm_unified 1.7.0rc1:
The lat_lon sets got error as follows:
```
Traceback (most recent call last):
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/run.py", line 34, in run_diags
    main(final_params)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/e3sm_diags_driver.py", line 334, in main
    index_path = create_viewer(path, parameters)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/viewer/main.py", line 128, in create_viewer
    result = viewer_function(root_dir, parameters)
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/viewer/default_viewer.py", line 184, in create_viewer
    cmip6_comparison_tuple = lat_lon_viewer.generate_lat_lon_cmip6_comparison(
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/viewer/lat_lon_viewer.py", line 396, in generate_lat_lon_cmip6_comparison
    test_model = read_e3sm_diags_metrics(
  File "/lcrc/soft/climate/e3sm-unified/base/envs/e3sm_unified_1.7.0rc1_chrysalis/lib/python3.9/site-packages/e3sm_diags/viewer/lat_lon_viewer.py", line 245, in read_e3sm_diags_metrics
    model = tmp[-6]
IndexError: list index out of range
2022-06-11 23:30:59,994 [INFO]: logger.py(move_log_to_prov_dir:91) >> Log file saved in model_vs_obs_1850-1851/prov/e3sm_diags_run.log
```
[...]

PR #599, opened 2022-06-13
Fix iss597 part2

chengzhuzhang:
Second PR for fixing #597:
- Change the logic to find model names in `read_e3sm_diags_metrics`
- Improve test name showing up in the legend box by including year range.


chengzhuzhang:
A complete run was tested on lcrc.

PR #619, opened 2022-08-31
Update branch for actions

forsyth2:
Update branch for actions. Since `master` was renamed `main`, we need to update the branch used for actions.

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -2,10 +2,10 @@ name: CI/CD Build Workflow
 
 on:
   push:
-    branches: [master]
+    branches: [main]
 
   pull_request:
-    branches: [master]
+    branches: [main]
 
   workflow_dispatch:
 
@@ -19,7 +19,7 @@ jobs:
     timeout-minutes: 2
     steps:
       - id: skip_check
-        uses: fkirc/skip-duplicate-actions@master
+        uses: fkirc/skip-duplicate-actions@main
         with:
           cancel_others: ${{ env.CANCEL_OTHERS }}
           paths_ignore: ${{ env.PATHS_IGNORE }}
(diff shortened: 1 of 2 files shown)


forsyth2:
@tomvothecoder The checks weren't successful and I can't merge this without them being successful. Is there something else that needs to be done? You also mentioned to add the bash commands to handle creating and deleting `gh-pages/index.html`.

.github/workflows/build_workflow.yml
```diff
@@ -152,7 +152,7 @@ jobs:
           git commit -m "Update documentation" -a || true
 
       - name: Push Changes
-        uses: ad-m/github-push-action@master
+        uses: ad-m/github-push-action@main
```
tomvothecoder, line 155:
```suggestion
        uses: ad-m/github-push-action@master
```
The branch reference here shouldn't be updated since it is for another repo.

.github/workflows/release_workflow.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           git commit -m "Update documentation" -a || true
 
       - name: Push Changes
-        uses: ad-m/github-push-action@master
+        uses: ad-m/github-push-action@main
```
tomvothecoder, line 64:
```suggestion
        uses: ad-m/github-push-action@master
```
The branch reference here shouldn't be updated since it is for another repo.

.github/workflows/build_workflow.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     timeout-minutes: 2
     steps:
       - id: skip_check
-        uses: fkirc/skip-duplicate-actions@master
+        uses: fkirc/skip-duplicate-actions@main
```
tomvothecoder, line 22:
```suggestion
        uses: fkirc/skip-duplicate-actions@master
```
The branch reference here shouldn't be updated since it is for another repo.

tomvothecoder (changes requested):
The build should work after my comments are addressed.

After this PR is merged, checkout the `gh-pages` branch and update `index.html` to reference `main`.

PR #620, opened 2022-08-31
update branch for actions

forsyth2:
Update branch for actions. Follow-up from #619.


forsyth2:
@tomvothecoder In `_build/html/main/index.html`, I changed the `master` references to `main`. I also renamed `_build/html/{master => main}`, which is the reason for the other 220 changes. 

I'm not sure why we have to change anything on the `gh-pages` branch though -- it seems like the actions on `main` should handle that.

tomvothecoder:
> @tomvothecoder In _build/html/main/index.html, I changed the master references to main. I also renamed _build/html/{master => main}, which is the reason for the other 220 changes.

You shouldn't need to do this, the new `main` docs should have already replaced those old `master` docs.

This isn't happening because because the build step for publishing docs is failing since it attempts to remove a non-existing directory (rm -r `_build/html/main`) before that directory is created. https://github.com/E3SM-Project/e3sm_diags/runs/8124895664?check_suite_focus=true

> I'm not sure why we have to change anything on the gh-pages branch though -- it seems like the actions on main should handle that.

I mentioned on Slack that creating `index.html` on `gh-pages` should be automated for cleaner workflows.

---

I'll try to quickly get this out of the way in the morning in another PR. You should be able to simply copy these updates over to other repos like `zstash` and `zppy`.

tomvothecoder:
This is superseded by #621.

Build is working again: https://github.com/E3SM-Project/e3sm_diags/runs/8140741213?check_suite_focus=true
Docs are working and published: [e3sm-project.github.io/e3sm_diags](https://e3sm-project.github.io/e3sm_diags)

PR #624, opened 2022-09-29
Add MERRA2 aerosol data for AOD diagnostics

### LINKED ISSUE #616: New Obs data: NASA MERRA2 assimilated AOD
Based on recommendation from @yfenganl : The MACv2 data is for 2005 only. Another data set to consider for aerosol climatology in future for E3SM diags is the NASA MERRA2 assimilated AOD ([https://gmao.gsfc.nasa.gov/research/science_snapshots/2015/MERRA2_global_aerosol_dist.php](https://urldefense.us/v3/__https://gmao.gsfc.nasa.gov/research/science_snapshots/2015/MERRA2_global_aerosol_dist.php__;!!G2kpM7uM-TzIFchu!kdsSs4kK1HO6l2NcO5JDOYAm-n-CQU3DxtvLQmgtQXXX2MIzA7CDobw2om2CRfZksw$)), which is updated to most recent years.

chengzhuzhang:
resolves #616 

- [x] Add pre-processing script, upload processed data to LCRC
- [x] Code modification to add new variables.


chengzhuzhang:
Hi @yfenganl, I have now added MERRA@ aerosol data in e3sm_diags as a new standard datasets in additional to MACv1 and MACv2. Here is the link to the viewer:https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/zhang40/MERRA2_AOD/viewer/lat_lon/index.html

and the metrics table:
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/zhang40/MERRA2_AOD/viewer/table-data/ANN_metrics_table.html

The global mean from MERRA2 is 0.14 which is higher than MACv1  and MACv2. The representing years are not the same. So I think it might still be useful to add an additional source of reference?

chengzhuzhang:
Hi @mingxuanwupnnl, would you also help look at the comment here: https://github.com/E3SM-Project/e3sm_diags/pull/624#issuecomment-1314603509
You brought up at one point that the high latitude should be masked out in MACv1. Since we now have newer datasets, maybe we can consider to retire MACv1 all together to avoid confusion?

yfenganl:
> The global mean from MERRA2 is 0.14 which is higher than MACv1 and MACv2. The representing years are not the same. So I think it might still be useful to add an additional source of reference?

Hi Jill, MERRA-2 assimilates multiple satellite (AVHRR, MODIS and MISR) and ground-based (AERONET) observations, different from MAC-v2, which uses the spatial information from the AeroCom I models and constrains the magnitude with AERONET and other sun photometers. It has been shown that the MERRA-2 AOD estimate (1998-2019) is higher than MAC-v2 (reference year 2005),  as generally satellite estimates > AERONET > models. Both datasets are within the range of AOD estimates from various modern datasets including models. 

So yeah, I also think that both provide good constraints, giving us a bounding box of AOD. Plus, we could use MERRA-2 to examine the aerosol trend in the 30+ years.

One question, I notice that you extracted the AOD time average (1980-2021) from the MERRA data. since we often calculate 30-year (1985-2014) or 15-year (2000-2014) climatology from E3SM, perhaps you could consider using one of these time averages from MERRA-2, e.g., 1985-2014? thanks for incorporating this dataset to the E3SM diags.

mingxuanwupnnl:
@chengzhuzhang Thanks for including me in the conversation. I agree that we can retire MACv1, considering the update in MACv2 and the global coverage in MACv2. I also agree with @yfenganl that we could consider using average for timer period 1985-2014 and 2000-2014 (or 2005-2014).

chengzhuzhang:
@yfenganl and @mingxuanwupnnl  Thank you both for your expert judgement!
Yan, It is reassuring that the variation of global mean values from multiple source are expected!
Mingxuan, okay, let's retiring MACv1, its source file is kind of legacy and there is a new version in place. 

Regarding to provide averaging of a different year range, I think it is a good idea. There is one configuration of e3sm_diags where one can input time-series files and set the start and end year for either model data or obs data. But this configuration is not known and practiced by many. I will consider to provide average of 1985-2014, as well as make the year-specific configuration example more accessible.

Issue #623, opened 2022-09-22
Bump to 2.7.1 (fix matplotlib rotation bug)


tomvothecoder:
Related to #622

## Why?
The build was breaking here: https://github.com/E3SM-Project/e3sm_diags/actions/runs/3109172215/jobs/5039383619#step:7:547

```python
ValueError: rotation must be 'vertical', 'horizontal' or a number, not 45
```

Source file: https://github.com/E3SM-Project/e3sm_diags/blob/ece5c28f618ada63ffe3c7c2024c4dcfecbd5718/e3sm_diags/plot/cartopy/area_mean_time_series_plot.py#L86

The latest version of matplotlib (3.6.0) expects the rotation argument to be a number, but is specified as a string.
This is the only plot that has rotation specified as string.
```python
        ax1.set_xticklabels(x_ticks_labels, rotation="45", fontsize=8)
```

I updated “45" to 45 and pushed the commit, so hopefully the build is successful in a bit.

## Solution
We will probably need to do a patch release on conda since we don’t have a constraint on `matplotlib`.

They changed the `set_rotation()` method between 3.5 and 3.6 to check validate the type.

This should probably be considered a breaking change, but idk if they follow semver.

## Additional Context
https://github.com/matplotlib/matplotlib/blob/v3.5.x/lib/matplotlib/text.py#L1171-L1182
```
    def set_rotation(self, s):
        """
        Set the rotation of the text.
        Parameters
        ----------
        s : float or {'vertical', 'horizontal'}
            The rotation angle in degrees in mathematically positive direction
            (counterclockwise). 'horizontal' equals 0, 'vertical' equals 90.
        """
        self._rotation = s
        self.stale = True
```

https://github.com/matplotlib/matplotlib/blob/v3.6.0/lib/matplotlib/text.py#L1180-L1199
```
    def set_rotation(self, s):
        """
        Set the rotation of the text.

        Parameters
        ----------
        s : float or {'vertical', 'horizontal'}
            The rotation angle in degrees in mathematically positive direction
            (counterclockwise). 'horizontal' equals 0, 'vertical' equals 90.
        """
        if isinstance(s, Real):
            self._rotation = float(s) % 360
        elif cbook._str_equal(s, 'horizontal') or s is None:
            self._rotation = 0.
        elif cbook._str_equal(s, 'vertical'):
            self._rotation = 90.
        else:
            raise ValueError("rotation must be 'vertical', 'horizontal' or "
[...]

PR #622, opened 2022-09-09
add example running with zppy and fix matplotlib rotation bug

chengzhuzhang:
Close #617. This example is useful for E3SM Diags manuscript as well.

PR #631, opened 2022-11-15
Bump to 2.8.0rc1

chengzhuzhang:
Bump to 2.8.0rc1

### DIFF
--- a/conda-env/prod.yml
+++ b/conda-env/prod.yml
@@ -8,5 +8,5 @@ dependencies:
   # =================
   - python=3.9.10
   - pip=22.0.3
-  - e3sm_diags=2.7.0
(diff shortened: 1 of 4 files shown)


tomvothecoder:
> @tomvothecoder it looks like to constrain matplotlib version is not required, since we already has fixed the breaking line in #622 ?

I believe you're correct. Since we fixed this issue, the conda recipe for v2.8.0rc1 does not need a constraint because it should be compatible with `matplotlib > 3.6` and backwards compatible with `matplotlib < 3.6`. I will close #623. 

This also looks good to merge, thanks.

tomvothecoder (approved):

chengzhuzhang:
Thank you! @tomvothecoder

PR #638, opened 2022-12-12
CDAT Migration Phase 1 - Replace`CDPParameter`

### LINKED ISSUE #633: CDAT Migration Phase 1 -  Replace `cdp_parameter.CDPParameter`
Related to #629 

Main Reference:

- [ ] core_parameter.py – class CoreParameter

Downstream:

- [ ]  annual_cycle_zonal_mean_parameter.py
- [ ]  area_mean_time_series_parameter.py
- [ ]  arm_diags_parameter.py
- [ ]  diurnal_cycle_parameter.py
- [ ]  enso_diags_parameter.py
- [ ]  lat_lon_land_parameter.py
- [ ]  lat_lon_river_parameter.py
- [ ]  meridional_mean_2d_parameter.py
- [ ]  qbo_parameter.py
- [ ]  streamflow_parameter.py
- [ ]  tc_analysis_parameter.py
- [ ]  time_series_parameter.py
- [ ]  zonal_mean_2d_parameter.py   
- [ ]  zonal_mean_2d_stratosphere_parameter.py

tomvothecoder:
Closes #633 

This PR removes `CDPParameter` as a dependency from the `CoreParameter` class and its sub-classes. This was a fairly non-invasive process. I tried to avoid refactoring too much code in this PR to avoid unwanted side-effects that would require more debugging (#639 is for refactoring).

The parameter classes are used the same way to configure `e3sm_diags`, which means there are no public API changes in this PR that warrant a major release.

### 1. Refactor `CoreParameter`
- [x] Remove `CoreParser` subclassing `CDPParameter` 
  -  `CDPParameter` is just a Python class that stores parameters as object attributes.
  - As a result, removing this dependency was straightforward and painless.
- [x] Port `__add_()` method from `CDPParameter`
  - This was the only method from `CDPParameter` that was being called in `e3sm_diags`.
  - I refactored the implementation so the logic is now simpler.

### 2. Refactor `CoreParameter` Inheritance
- [x] Understand how `CoreParameter` class is sub-classed and extended
  - Each child parameter subclass (e.g. `ZonalMean2dParameter`) either does nothing except sub-class, modifies or updates attributes, adds new attributes, or has unique `check_values()`
  - `SETS_TO_PARAMETER` defined in `parameter/__init__.py` maps the string name of "set" to the parameter class
- [x] Figure out an extensible way of inheriting from `CoreParameter` for each diagnostic set

**UPDATE: I decided to keep the current design pattern to minimize side-effects. Opened up a new issue to address refactoring here #639.**

### 3. Testing and cleanup
- [x] Add unit tests for each parameter class 
  - The only method that I did not test was `TimeSeriesParameter.check_values()` because the conditional statements are a bit too complex for me to sit down and understand for this PR. I added a TODO statement to revisit this later.
- [x] Test locally with `run_e3sm_diags.py` script
- [x] Integration tests still pass

### DIFF
--- a/e3sm_diags/driver/utils/dataset.py
+++ b/e3sm_diags/driver/utils/dataset.py
@@ -281,7 +281,10 @@ class Dataset:
         path = self.parameters.reference_data_path
         data_name = self.parameters.ref_name
 
-        if hasattr(self.parameters, "ref_file"):
+        if (
+            hasattr(self.parameters, "ref_file")
+            and getattr(self.parameters, "ref_file") != ""
+        ):
             fnm = os.path.join(path, self.parameters.ref_file)
             if not os.path.exists(fnm):
                 raise IOError("File not found: {}".format(fnm))
--- a/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
+++ b/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
@@ -6,18 +6,19 @@ from e3sm_diags.parameter.zonal_mean_2d_stratosphere_parameter import (
     ZonalMean2dStratosphereParameter,
 )
 
-if TYPE_CHECKING:
-    from e3sm_diags.parameter.core_parameter import CoreParameter
-
 
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     """Creates the mean, max, min, rmse, corr in a dictionary"""
     return base_create_metrics(ref, test, ref_regrid, test_regrid, diff)
 
 
+if TYPE_CHECKING:
+    from e3sm_diags.parameter.zonal_mean_2d_parameter import ZonalMean2dParameter
+
+
 def run_diag(
     parameter: ZonalMean2dStratosphereParameter,
-) -> CoreParameter:
+) -> ZonalMean2dParameter:
     return base_run_diag(
         parameter, default_plevs=ZonalMean2dStratosphereParameter().plevs
     )
--- a/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
+++ b/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from typing import TYPE_CHECKING
 
 from e3sm_diags.driver.zonal_mean_2d_driver import create_metrics as base_create_metrics
 from e3sm_diags.driver.zonal_mean_2d_driver import run_diag as base_run_diag
@@ -6,13 +6,18 @@ from e3sm_diags.parameter.zonal_mean_2d_stratosphere_parameter import (
     ZonalMean2dStratosphereParameter,
 )
 
+if TYPE_CHECKING:
+    from e3sm_diags.parameter.core_parameter import CoreParameter
+
 
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     """Creates the mean, max, min, rmse, corr in a dictionary"""
     return base_create_metrics(ref, test, ref_regrid, test_regrid, diff)
 
 
-def run_diag(parameter):
+def run_diag(
+    parameter: ZonalMean2dStratosphereParameter,
+) -> CoreParameter:
     return base_run_diag(
         parameter, default_plevs=ZonalMean2dStratosphereParameter().plevs
     )
--- a/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
+++ b/e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
@@ -1,7 +1,6 @@
-from typing import TYPE_CHECKING
-
 from e3sm_diags.driver.zonal_mean_2d_driver import create_metrics as base_create_metrics
 from e3sm_diags.driver.zonal_mean_2d_driver import run_diag as base_run_diag
+from e3sm_diags.parameter.zonal_mean_2d_parameter import ZonalMean2dParameter
 from e3sm_diags.parameter.zonal_mean_2d_stratosphere_parameter import (
     ZonalMean2dStratosphereParameter,
 )
@@ -10,10 +9,6 @@ from e3sm_diags.parameter.zonal_mean_2d_stratosphere_parameter import (
 def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     """Creates the mean, max, min, rmse, corr in a dictionary"""
     return base_create_metrics(ref, test, ref_regrid, test_regrid, diff)
-
-
-if TYPE_CHECKING:
-    from e3sm_diags.parameter.zonal_mean_2d_parameter import ZonalMean2dParameter
 
 
 def run_diag(
--- a/e3sm_diags/e3sm_diags_driver.py
+++ b/e3sm_diags/e3sm_diags_driver.py
@@ -351,7 +351,7 @@ def _run_with_dask(parameters: List[CoreParameter]) -> List[CoreParameter]:
     https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.compute.html
     """
     bag = dask.bag.from_sequence(parameters)
-    config = {"scheduler": "processes", "context": "fork"}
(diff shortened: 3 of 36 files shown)


tomvothecoder (commented):
Initial review comments

e3sm_diags/parameter/core_parameter.py
```diff
@@ -33,68 +75,129 @@ def __init__(self):
...
+
+        for attr in dir(other):
+            # Ignore any of the hidden attributes.
+            if attr.startswith("_") or isinstance(getattr(other, attr), MethodType):
+                continue
+
+            val = copy.deepcopy(getattr(other, attr))
+            setattr(duplicate, attr, val)
+
+        return duplicate
```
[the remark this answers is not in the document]
tomvothecoder: This method has been refactored.

e3sm_diags/parameter/core_parameter.py
```diff
@@ -33,68 +75,129 @@ def __init__(self):
...
+        # The default values for these attributes are set to "" in `__init__`.
+        # FIXME: The user should pass these values into the object
+        # initialization rather than having to check they are set here.
+        must_have_params = ["reference_data_path", "test_data_path", "results_dir"]
         for param in must_have_params:
-            if not hasattr(self, param):
+            if getattr(self, param) != "":
                 msg = "You need to specify {p} in the parameters file or via the command line using --{p}".format(
                     p=param
                 )
```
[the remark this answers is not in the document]
tomvothecoder: This check was fixed to `if getattr(self, param) == ""`

e3sm_diags/parameter/core_parameter.py
```diff
@@ -1,18 +1,59 @@
-from __future__ import print_function
+import copy
+from typing import Dict, List, Optional
 
-import cdp.cdp_parameter
 
-
-class CoreParameter(cdp.cdp_parameter.CDPParameter):
```
tomvothecoder, line 6: Most of the "important" changes are in the `core_parameter.CoreParameter` class.

Changes include:
 - Removing `CDPParameter` sub-classing
 - Organize attributes with comments
   - Adding missing attributes that are set on `CoreParameter` objects in various locations of the codebase
   - Add type annotations for attributes and variables
 - Add `__add__` which is ported from `CDPParameter`
- Update `check_values()` and its `must_have_params` list

e3sm_diags/parameter/diurnal_cycle_parameter.py
```diff
@@ -4,8 +4,8 @@
 class DiurnalCycleParameter(CoreParameter):
     def __init__(self):
         super(DiurnalCycleParameter, self).__init__()
+        # Custom attributes
+        # =============================
+        self.normalize_amp_int = 0
         self.print_statements = False
         self.normalize_test_amp = False
-        self.ref_timeseries_input = False
-        self.test_timeseries_input = False
```
tomvothecoder, line 10: Remove `self.ref_timeseries_input = False` and `self.test_timeseries_input = False` because they are both default to `False` in parent `CoreParameter` class.

e3sm_diags/parameter/zonal_mean_2d_parameter.py
```diff
@@ -8,35 +8,9 @@
...
-        #    750.0,
-        #    800.0,
-        #    850.0,
-        #    875.0,
-        #    900.0,
-        #    925.0,
-        #    950.0,
-        #    975.0,
-        #    1000.0,
-        # ]
```
tomvothecoder, line 39: Remove commented out code.

tomvothecoder (commented):
Hi @chengzhuzhang, this PR is ready for review at your convenience. I suggest using the [Files Changes](https://github.com/E3SM-Project/e3sm_diags/pull/638/files) tab because I made two separate self-reviews.

I left review comments for you to look at it. Most of the changes in `driver/` files are just type annotations.

chengzhuzhang:
Hi @tomvothecoder thank you for the PR and the careful comments for a review. The code changes look good to me. I will test a full run and update here to make sure the changes won't break any sets.

e3sm_diags/driver/utils/dataset.py
```diff
@@ -281,7 +281,10 @@ def get_ref_filename_climo(self, season):
         path = self.parameters.reference_data_path
         data_name = self.parameters.ref_name
 
-        if hasattr(self.parameters, "ref_file"):
+        if (
```
chengzhuzhang, line 284: Add change to reconcile with default "ref_file" setting.

e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
```diff
@@ -12,7 +11,9 @@ def create_metrics(ref, test, ref_regrid, test_regrid, diff):
     return base_create_metrics(ref, test, ref_regrid, test_regrid, diff)
 
 
-def run_diag(parameter):
+def run_diag(
+    parameter: ZonalMean2dStratosphereParameter,
+) -> ZonalMean2dParameter:
```
chengzhuzhang, line 16: @tomvothecoder I'm not exactly sure why using `coreParameter` here triggers error "NameError: name 'CoreParameter' is not defined" at the first place and  then `mypy` insist to expect  ZonalMean2dParameter. But things run okay after my changes.

tomvothecoder: Looks good with this change.

For your error:
---
You need to add `from __future__ import annotations` to reference the `CoreParameter` module without quotes, otherwise you get that `NameError`.

Without `import annotations`:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from e3sm_diags.parameter.core_parameter import CoreParameter

def run_diag(
	parameter: ZonalMean2dStratsophereParameter
) -> "CoreParameter"
```

With `import annotations`:
```
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from e3sm_diags.parameter.core_parameter import CoreParameter

def run_diag(
	parameter: ZonalMean2dStratsophereParameter
) -> CoreParameter
```
This is a "[postponed evaluation of annotations](https://peps.python.org/pep-0563/)", which I believe means evalutation only happens through a typechecker like `mypy` and not at runtime. It is also cleaner without quotes in my opinion.

e3sm_diags/driver/utils/dataset.py
tomvothecoder: Thanks!

tomvothecoder:
Hi @chengzhuzhang, let me know if you are finished testing and if it is ready to be merged :)

e3sm_diags/driver/zonal_mean_2d_stratosphere_driver.py
chengzhuzhang: Thanks! Good to learn this!

chengzhuzhang:
> Hi @chengzhuzhang, let me know if you are finished testing and if it is ready to be merged :)

Thanks for addressing my questions. The full-set standard model vs obs run is all good now. I'm gonna test a couple of corner cases now, will let you when its ready.

e3sm_diags/e3sm_diags_driver.py
```diff
@@ -351,7 +351,7 @@ def _run_with_dask(parameters: List[CoreParameter]) -> List[CoreParameter]:
     https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.compute.html
     """
     bag = dask.bag.from_sequence(parameters)
-    config = {"scheduler": "processes", "context": "fork"}
+    config = {"scheduler": "processes", "multiprocessing.context": "fork"}
```
chengzhuzhang, line 354: The freeze_support() issue came back in one of my test cases. When replace with the native keyword `multiprocessing.content`, it solved the problem.

chengzhuzhang (approved):

chengzhuzhang:
Hey Tom, I had a final fix. Now my tests all run okay. Please review and feel free to merge if it looks fine with you!

tomvothecoder:
Thanks @chengzhuzhang! Merging now.

PR #676, opened 2023-02-28
fix typo in aerosol budget printout

mahf708:
Fixing a typo in the aerosol budget table. Sinks should have units of Tg/yr (as the rest). Marking as draft as I don't know if you accept PRs. Feel free to cherry pick or fix as you see fit.

### DIFF
--- a/auxiliary_tools/aerosol_budget.py
+++ b/auxiliary_tools/aerosol_budget.py
@@ -51,7 +51,7 @@ def generate_metrics_dic(data, aerosol, season):
     print(f'{aerosol} Lifetime (days): ',f'{burden_total/sink*365:.3f}')
     metrics_dict = {
     "Surface Emission (Tg/yr)": f'{srfemis:.3f}',
-    "Sink (Tg/s)": f'{sink:.3f}',
+    "Sink (Tg/yr)": f'{sink:.3f}',
     "Dry Deposition (Tg/yr)": f'{drydep:.3f}',
     "Wet Deposition (Tg/yr)": f'{wetdep:.3f}',
     "Burden (Tg)": f'{burden_total:.3f}',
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
Nice catch! Thank you @mahf708. I will review once the PR is ready.

chengzhuzhang:
Thanks for the sanity check @mahf708 It is concerning that derived global mean from 3D mass concentration data would deviate from the BURDENxx output so much. And more mysterious the relative difference changes with species. Other than difference in the column integration and global averaging. Could it be that there is missing terms in these burden output? Also, I don't seem to find "BURDENBC", "BURDENDUST", "BURDENMOM", "BURDENSOA", "BURDENSO4", "BURDENPOM", "BURDENSEASALT" in standard output.

chengzhuzhang:
@mahf708 Thanks for your valuable input! [Here](https://github.com/E3SM-Project/e3sm_diags/blob/833d9b2df0515565bf4150b20101e80dea647294/e3sm_diags/driver/aerosol_budget_driver.py#L29) is how column integration is done. The alternative you suggested should be straightforward to implementation, but I still hope to understand if there are discrepancies between 2d BURDEN terms and derived from 3d Mass  other than the vertical integration difference. It would be the best to use the same source terms when compare across models.

chengzhuzhang:
@mahf708 I think at the last aersol working group call, with Mingxuan's new PR, I think your simulation have both BURDEN_ and Mass_* variables. Would you share your path for climatology data, and I can try do a comparison.

tangq:
@mahf708 , this column integration example is for total column ozone (TOZ), and aerosol's should be similar except the units conversion. Look at lines 1057-1062.

https://github.com/E3SM-Project/v3atm/blob/NGD_v3atm/components/eam/src/chemistry/mozart/mo_chm_diags.F90#L1063

chengzhuzhang:
Results show 2d BURDEN* output versus 3d MASS* using e3sm_diags using @mahf708's run: F2010.PD.NGD_v3atm.0096484.compy and ANN average. 


e3sm_diags* | BURDEN* | MASS*
-- | -- | --
BC (Tg) | 0.199 | 0.202
DUST (Tg) | 31.144 | 31.663
MOM (Tg) | 0.073 | 0.079
SOA (Tg) | 5.082 | 5.467
SO4 (Tg) | 0.999 | 1.015
POM (Tg) | 1.818 | 1.864
SEASALT (Tg) | 2.917 | 2.970

There does seem to be discrepancies between two methods. Consistent with Naser's results, using MASS* output always gives slightly higher numbers, and for some species the difference could be large, in this case SOA numbers deviates about 7%.  It is expected that there will be difference due to that the BURDEN* case, use the model column integration routine,  the burden is computed in e3sm_diags with MASS* output. It remains unclear if all components of each species are taken in account for both BURDEN and MASS output.

chengzhuzhang (approved):
Thanks for the contribution @mahf708 ! I have documented the difference between using BURDEN output and MASS. At this point, I'm not sure if it is reliable to switch between these two group of output based on availability. I will merge this PR and keep an eye out if the discrepancies can be explained.

chengzhuzhang:
I was particularly interested in the sea salt difference... In any case, the PR is merged! Thanks again for all the input and your kind words.

Issue #664, opened 2023-02-21
CDAT Migration Phase 2: Refactor `qbo` set


tomvothecoder:
Todo:
- [ ]  [Refactor `VarHandler._get_attr_from_climo()`](https://github.com/E3SM-Project/e3sm_diags/pull/677#discussion_r1336444331)

tomvothecoder:
Closed by #826

Issue #165, opened 2018-06-01
Unstructured grid files are read for lat-lot set by mistake


oksanaguba:
On Edison, I have 3 very similar runs. They all have the same duration, output, etc. However, initially diagnostics with lat-lon was producing different plots for all 3. Many plots were missing. For example, all ANN plots were missing from 2 of 3 runs, while all 3 runs have the same data structures.

Turns out that ncclimo produces files with structured (latlon) and unstructured grids (ncol). I did not know which are needed, so, I pull all of them in the folder where diagnostics was running. However, diagnostics would pick different files for observations for different runs. It would open file with unstructured grid for one run and file with structured grid for another. Turns out that in function _findfile(), value

listdir_files = os.listdir(path_name)

was used to find the first match for a file. The first match is not necessarily a structured file, how it should be (and listdir_files does not seem to have any particular order). In the function, filename match has to satisfy case name and season, but nothing else. In 2 out of 3 runs, first match would be a file with unstructured grid, and diagnostics would produce an error, no plot. 

The best solution for me was to move all unstructured files (the ones with mask `*000*_climo.nc`) . There are still some errors in the output (at least some of them are related to non existing fields in observations), but at least sets of diagnostics plots now match for my 3 runs, as supposed to.

Initially, the error about unstructured files looked like this
```
test file: /scratch1/scratchdirs/onguba/vsplit-diag/may09-f2/vsplit-f2-6y_ANN_000112_000611_climo.nc
reference file: /global/project/projectdirs/acme/acme_diags/obs_for_acme_diags/GPCP_v2.2_ANN_climo.nc
Variable: PRECT
Selected region: global
no domain selector
Error in acme_diags.driver.lat_lon_driver
Traceback (most recent call last):
  File "/global/homes/o/onguba/.conda/envs/e3sm_diags_env/lib/python2.7/site-packages/acme_diags/acme_diags_driver.py", line 79, in run_diag
    single_result = module.run_diag(parameters)
  File "/global/homes/o/onguba/.conda/envs/e3sm_diags_env/lib/python2.7/site-packages/acme_diags/driver/lat_lon_driver.py", line 235, in run_diag
    mv1_domain, mv2_domain, parameter.regrid_tool, parameter.regrid_method)
  File "/global/homes/o/onguba/.conda/envs/e3sm_diags_env/lib/python2.7/site-packages/acme_diags/driver/utils.py", line 164, in regrid_to_lower_res
[...]

PR #677, opened 2023-03-17
CDAT Migration Phase 2: Refactor core utilities and  `lat_lon` set

### LINKED ISSUE #658: CDAT Migration Phase 2: Refactor `lat_lon` set
/## Overview
The components to refactor include the driver, the plotter, and the viewer. Each file has a set of functions to refactor and module references that might need to be refactored as well.

We will use the ["bubble context"](https://philippe.bourgau.net/incremental-software-development-strategies-for-large-scale-refactoring-number-4-a-pattern-language/#bubble-context) refactoring method for this task.

### File 1 - `lat_lon_driver.py`

Order of function calls:  `run_diag()` -> `create_metrics()` -> `create_and_save_data_and_metrics()`

- [x] Refactor `utils.dataset.Dataset` class methods to operate on xr.Dataset/xr.DataArray objects
   a. Create skeleton methods for methods that need to be factored (append `_new` to name)
   b. Add failing unit tests for these methods
   c. Implement new methods
   d. Update references from old methods to new methods
- [x] Refactor general utilities that are called by the driver to operate on xarray objects (e.g., `climo.py`)
   a. Create a `utils_new.py`
   b. Add skeleton function definitions for new general utilities
   c. Add failing unit tests for these functions
   d. Implement new general utilities
   e. Update references from old utilities to new utilities
- [x] Refactor `create_and_save_data_and_metrics()` and `create_metrics()` 
   a. These functions call `e3sm_diags.metrics` functions

#### Module References - [tree diagram](https://drive.google.com/file/d/1U7d1nwTNWJ9DQPEcp9x57AUmCu1muL3D/view?usp=sharing)

These are the other modules referenced in `lat_lon_driver.py` that need to be refactored to operate on xr.DataArray/xr.Dataset objects.

- `e3sm_diags.utils`
  - [x] `dataset_new.py` (`class Dataset`) - Refactor methods that call cdms2 (e.g., `cdms2.open()` )
     - [x] `get_static_variable()`
     - [x] `get_attr_from_climo()`
     - [x] `get_climatology_variable()` 
       - [x] `_get_climo_var()`
        - [x] `get_timeseries_var()`
        - [x] `_get_var_from_timeseries_file()`
        - [x] `e3sm_diags/driver/utils/climo.py` -> create `climo_xr.py`
  - [x] `io.py` (replaces I/O functions in `general.py`)
    - [x] `get_output_dir()`
    - [x] `save_ncfiles()`
    - [x] `get_name_and_yrs()` -- added as a method to `Dataset` class
  - [x] `regrid.py` (replaces regridding functions in `general.py`)
    - [x] (NEW) `has_z_axis` (replaces `cdms2.axis.getLevel()`)
    - [x] `regrid_to_lower_res()` - 100% identical results
    - [x] `convert_to_pressure_levels()` -- Really close (1e-6 max abs and 1e-8 max rel diffs) 
    - [x] `select_region()` -- Really close (1e-7 max abs and rel diffs without land/sea region masking lower limit)
      - ~~Opened an xcdat [GitHub discussion post](https://github.com/xCDAT/xcdat/discussions/521) to figure out why xesmf and cdms2 esmf regridder produces different regridded land sea masks~~ -- caused by cdms2 not importing esmf properly, has since been patched 
      - Replaced by `_apply_land_sea_mask()` and `_subset_on_region()`
- `e3sm_diags.metrics_xr.py` (replaces `metrics/__init__.py`)
  - [x] `corr` -- virtually identical (1e-14 max abs and rel diffs)
  - [x] `mean` - virtually identical (1e-12 max abs and 1e-14 max rel diffs)
  - [x] `rmse` - virtually identical (1e-13 max abs and 1e-14 max rel diffs)
  - [x] `std` - virtually identical (1e-14 max abs and 1e-15 max rel diffs)

#### Functions to Refactor
- `e3sm_diags.plot.plot()` 
  - [x] `_get_plot_fnc()`
    - [x] `lat_lon_plot.plot()`  -- moved sub-functions below to `plot/utils.py`
      - [x] `plot_panel()`
      - [x] `determine_tick_step()`
      - [x] `get_ax_size()`
      - [x] `add_cyclic()`

#### Module References
- `e3sm_diags.derivations.default_regions` - `region_specs`
- `e3sm_diags.driver.utils.general` - `get_output_dir()`
- `e3sm_diags.plot` - `get_color_map()`

#### Functions to Refactor
- `create_viewer()` -> `seasons_used()`, `_get_description()`, `_add_to_lat_lon_metrics_table()`, `create_metadata()`, and `_add_information_to_viewer()`

#### Module References
- `e3sm_diags.parser` - `SET_TO_PARSER`
- `e3sm_diags.viewer.utils` - `add_header()`, `h1_to_h3()`
- `e3sm_diags.viewer.lat_lon_viewer` - `generate_lat_lon_metrics_table()`, `generate_lat_lon_taylor_diag()`, `generate_lat_lon_cmip6_comparison()`, `generate_lat_lon_metrics_table()`

tomvothecoder:
## Description

- Closes #658 

This PR refactors most core utilities for e3sm_diags to use Xarray/xCDAT since CDAT is deprecated at the end of the year. The `lat_lon` set is also refactored to use these core utilities. 

## How to Review

This is a large PR that involves many changes. I am interested in code review for areas of improvement, optimization, potential bugs, etc. The functionality itself is unchanged and working _(refer to viewer outputs at the bottom of the description)_. 

To help streamline the review process, there is section below for the main focus areas. If you are interested and have time, feel free to review other files and functions that are not listed. 

**I anticipate review will take the next few weeks (targeting end of Sep/2023)**. I'll try my best to respond to any comments, suggestions, or questions you might have. Thanks for the help!

### **Main Focus Areas**

Each focus area includes the most important files and functions to review. 
* Most refactored functions include a docstring with a "Notes" section at the bottom. The "Notes" section lists the old function it replaces (if you wanted to compare them). Here's an example (look at the bottom of the docstring):
* https://github.com/E3SM-Project/e3sm_diags/blob/3e0a0871736f89b9bb24968b25c58539f54ce975/e3sm_diags/driver/utils/regrid.py#L180-L200

#### (1) Dataset class 
- `driver/dataset_xr.py`
  - Methods have been reordered to make the class more readable
  - Methods have been renamed, refactored, and updated with docstrings. I tried to group related time series and climatology functions together.
  - Most methods now return `xr.Dataset` objects, which are crucial because they contain bounds for downstream operations (e.g., generating weights). Variables in `xr.Dataset` objects should be operated on (regridding, regional subsetting, metrics calculation) until the last step, plotting (which will use the variables directly as `xr.DataArray`).

#### (2) Computation, Regridding, and I/O Utilities
- `driver/utils/climo_xr.py` 
  - `climo()` function has the same logic as the older version but operates on Xarray objects
- `driver/utils/regrid.py`
  - `align_grids_to_lower_res` (formerly `regrid_to_lower_res()`)
  - `regrid_z_axis_to_plevs()` (formerly `convert_to_plevs()`)
    - `_hybrid_to_plevs()`
    - `_pressure_to_plevs()`
- `driver/metrics/metrics.py` 
  - `spatial_avg()` (formerly `mean()`)
  - `std()`
  - `correlation()`
  - `rmse()`
- `driver/utils/io.py`
  - Moved `_write_vars_to_netcdf()` and `_get_output_dir()` from `general.py` to this file

#### (3) Lat Lon Driver 
- _`driver/lat_lon_driver.py`_
  - Moved `run_diag()` to the top level of the file
  - Most of the code is similar. I replaced function calls to use the new refactored functions, renamed things, and tried to refactor logic to make it more readable.
  - `xr.Dataset` objects are mainly used because they contain bounds. The test/reference variable gets updated in the `xr.Dataset` objects when applying land sea mask, regridding, etc. These objects are passed to `_create_and_save_metrics()`.

#### (4) Lat Lon Plotter
- _`plot/lat_lon_plot.py`_
  - Refactored arguments to expect `xr.DataArray` type
  - Extract code into separate utility functions (added to `plot/utils.py`), which reduced the length of the file significantly (compare to `plot/cartopy/deprecated_lat_lon_plot.py`)

#### (5) Derived Variables 
- _`derivations/acme.py`_
  - Split `acme.py` file into `derivations.py`, `formulas.py`, and `utils.py` 
  - `derivations.py` -- stores derived variables dictionary. 
  - `derivations/formulas.py`  -- stores formulas for deriving variables, all functions have been updated to operate on `xr.DataArray` objects
  - `utils.py`  -- a general file for storing utility functions (e.g,. unit conversion)
    - `_apply_land_sea_mask()` replaces the old `mask_by()` function (removed unnecessary logic and renamed for clarity)
    - _TODO: Some functions have not been refactored (e.g., `determine_tau()`)_
- _`derivations/default_regions_xr.py`_
  - Updated structure of `REGION_SPECS` dict (formerly `region_specs`)
  - Updated structure of `ARM_SITE_SPECS` dict (formerly `points_specs`)

#### (6) Misc.
- `e3sm_diags/driver/utils/general.py`
  - Removed the `mask_by()` function because it was overly complex and `select_region()` only uses one part of the code.
- `e3sm_diags/driver/utils/core_parameter.py`
  - Updated `self.backend="mpl"` to `self.backend=cartopy` because "mpl" gets updated to "cartopy" in `e3sm_diags/plot/__init__.py` anyways

## Checklist

If applicable:

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

### Viewer Outputs

- This branch: https://portal.nersc.gov/cfs/e3sm/vo13/examples_658/ex1_modTS_vs_modTS_3years/viewer/lat_lon/index.html
- `main` branch: https://portal.nersc.gov/cfs/e3sm/vo13/examples/ex1_modTS_vs_modTS_3years/viewer/lat_lon/index.html

### Future Items
- #734


e3sm_diags/driver/utils/climo.py
```diff
@@ -45,6 +45,10 @@ def climo(var, season):
     v = var.asma()
 
     # Compute climatology
+    # FIXME: season can be "ANN" and there is no conditional for "ANN", which
+    # breaks this function. The old Dataset class would call this function
+    # and drivers would do a try and except that captures Exception (all
+    # exceptions), which produces a silent error.
     if season == "ANNUALCYCLE":
         cycle = [
             "01",
```
tomvothecoder, line 54: Just a note that the old `climo()` function breaks when `season == "ANN"` because there is not conditional statement to set `cycle`.

chengzhuzhang:
> Hi @chengzhuzhang, I'm spending more time managing the effort and progress for this PR.
> 
> Here are a few notes:
> 
>     1. This is a large and high priority task so we might need to pull another person in to help refactor other components if we want it merged within the next 1-2 months. xCDAT and other projects are currently taking up a lot of my bandwidth.
>        
>        * **Workaround -- I refactor the core driver and utilities, and somebody else refactors the plotter and viewer**
> 
>     2. I'm sinking too much time on investigating the diffs between xCDAT's climatology vs. E3SM diags climatology
>        
>        * **Workaround -- Refactor E3SM diags climatology to operate on xr.DataArray instead of TransientVariable. Open up a new issue to integrate xCDAT's climatology later, and attach my investigation Jupyter Notebook**
> 
>     3. I don't have a clear timeline on when Jason can work on incorporating a version of [cdutil.reconstructPressureFromHybrid](https://github.com/CDAT/cdutil/blob/b823b69db46bb76536db7d435e72075fc3975c65/cdutil/vertical.py#L8) into `cf-xarray`.
>        
>        * **Temporary Workaround -- Create a function in e3sm_diags that is a copy of this code and convert it to xarray directly**

Hey @tomvothecoder thank you for the thoughtful suggestions. This is a high priority PR, since we have a crunching time line for the CDAT support. The workaround for climatology and and pressure level construction sounds good to me! Both @forsyth2  and I are familiar with the plotter and viewer code, so I think distribute the refactor work to either would make sense. We can talk about how to get start splitting the work.

e3sm_diags/driver/utils/climo.py
tomvothecoder: The comment above is incorrect. There is a conditional for "ANN".

pyproject.toml
```diff
@@ -25,3 +26,30 @@ exclude = '''
...
+line_length = 88
+skip = "e3sm_diags/e3sm_diags_driver.py"
+
+[tool.pytest.ini_options]
+# Docs: https://docs.pytest.org/en/7.2.x/reference/customize.html#configuration
+junit_family = "xunit2"
+addopts = "--cov=e3sm_diags --cov-report term --cov-report html:tests_coverage_reports/htmlcov --cov-report xml:tests_coverage_reports/coverage.xml -s --ignore=tests/integration"
+python_files = ["tests.py", "test_*.py"]
+testpaths = "tests/e3sm_diags"
+filterwarnings = "ignore:.*yaksa"
```
tomvothecoder, line 45:
```suggestion
[tool.pytest.ini_options]
# Docs: https://docs.pytest.org/en/7.2.x/reference/customize.html#configuration
junit_family = "xunit2"
addopts = "--cov=e3sm_diags --cov-report term --cov-report html:tests_coverage_reports/htmlcov --cov-report xml:tests_coverage_reports/coverage.xml -s"
python_files = ["tests.py", "test_*.py"]
testpaths = "tests/e3sm_diags"
```

tomvothecoder:
Hey @chengzhuzhang and @forsyth2, this PR is ready for code review. The PR description includes all of the information needed to help make the review process smooth since it's a large PR.

I have a few items left including validating the metrics for closeness compared to `main`, fixing the unit tests, and making sure integration tests pass. After everything checks out, I will merge this PR into a `cdat-refactor-fy23` development branch, which we will merge into `main` periodically until all sets are refactored.

chengzhuzhang:
> Hey @chengzhuzhang and @forsyth2, this PR is ready for code review. The PR description includes all of the information needed to help make the review process smooth since it's a large PR.
> 
> I have a few items left including validating the metrics for closeness compared to `main`, fixing the unit tests, and making sure integration tests pass. After everything checks out, I will merge this PR into a `cdat-refactor-fy23` development branch, which we will merge into `main` periodically until all sets are refactored.

Super! Thank you for the notes. I will start review soon!

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+def align_grids_to_lower_res(
+    ds_a: xr.Dataset,
+    ds_b: xr.Dataset,
+    var_key: str,
+    tool: REGRID_TOOLS,
+    method: str,
+) -> Tuple[xr.DataArray, xr.DataArray]:
+    """Align the grids of two Dataset using the lower resolution of the two.
+
+    A variable has a lower resolution if it has less latitude coordinates,
```
forsyth2, line 245: "less" => "fewer"

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+
+    References
+    ----------
+    .. [1] https://xcdat.readthedocs.io/en/stable/generated/xarray.Dataset.regridder.horizontal.html
+    """
+    # TODO: Accept "esmf" as `tool` value for now because `CoreParameter`
+    # defines `self.regrid_tool="esmf"` by default and
+    # `e3sm_diags.driver.utils.general.regrid_to_lower_res()` accepts "esmf".
+    # Once this function is deprecated, we can remove "esmf" as an option here
+    # and update `CoreParameter.regrid_tool` to "xesmf"`.
```
forsyth2, line 291: Does this `TODO` still apply?

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+
+
+def regrid_z_axis_to_plevs(
+    dataset: xr.Dataset,
+    var_key: str,
+    plevs: List[int] | List[float],
+) -> xr.Dataset:
+    """Regrid a variable's Z axis to the desired pressure levels (mb units).
+
+    The Z axis (e.g., 'lev') must either by hybrid or pressure, which is
```
forsyth2, line 327: So is "isobaric" valid or not?

forsyth2: Is "isobaric" just a synonym for "pressure"??? And in that case, what is "hybrid" a hybrid of?

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+    -------
+    xr.Dataset
+        The variable with a Z axis regridded to pressure levels (mb units).
+
+    Notes
+    -----
+    Replaces `e3sm_diags.driver.utils.general.hybrid_to_plevs`.
+    """
+    # TODO: Do we need to convert the Z axis to mb units if it is in PA? Or
+    # do we always expect units to be in mb?
```
forsyth2, line 426: This looks like it is still `TODO`?

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+    ds = dataset.copy()
+
+    z_axis, _ = xc.create_axis("lev", plevs, generate_bounds=False)
+    pressure_grid = xc.create_grid(z=z_axis)
+
+    pressure_coords = _hybrid_to_pressure(ds, var_key)
+
+    # Make sure that the input dataset as Z axis bounds, which are required for
```
forsyth2, line 434: "as" => "has"??

e3sm_diags/metrics/metrics.py
```diff
@@ -0,0 +1,233 @@
...
+
+def correlation(
+    da_a: xr.DataArray,
+    da_b: xr.DataArray,
+    weights: xr.DataArray,
+    serialize: bool = False,
+) -> xr.DataArray:
+    """Compute the correlation coefficient between two variables.
+
+    This function uses the Pearson correlation coefficient. Refer to [1]_ for
```
forsyth2, line 125: Is the underscore supposed to be there? Does that create a footnote in the online docs?

e3sm_diags/driver/utils/io.py
```diff
@@ -0,0 +1,98 @@
...
+    ref : xr.DataArray
+        The reference variable.
+    diff : Optional[xr.DataArray]
+        The optional difference between the test and reference variables.
+
+    Notes
+    -----
+    Replaces `e3sm_diags.driver.utils.general.save_ncfiles()`.
+
+    # TODO: Do we need to consider the append option like in save_ncfiles()?
```
forsyth2, line 39: It looks like this still needs to be addressed?

e3sm_diags/driver/utils/io.py
```diff
@@ -0,0 +1,98 @@
...
+        If the directory does not exist and could not be created.
+    """
+    results_dir = parameter.results_dir
+    dir_path = os.path.join(results_dir, parameter.current_set, parameter.case_id)
+
+    # Create the directory if it does not exist. When running diags in parallel,
+    # a process for another set can create the dir already so we can ignore
+    # creating the dir for this set.
+    if not os.path.exists(dir_path):
+        # TODO: Write a unit test for this try/except statement.
```
forsyth2, line 91: Is this done?

e3sm_diags/driver/utils/20230906-153411-issue-658-metrics-diff.xlsx
forsyth2, line 1: Is this binary file necessary for something?

e3sm_diags/derivations/derivations.py
```diff
@@ -0,0 +1,1484 @@
...
+                ("PRECSC", "PRECSL"),
+                lambda precsc, precsl: precst(precsc, precsl),
+            ),
+        ]
+    ),
+    # Sea Surface Temperature: Degrees C
+    # Temperature of the water, not the air. Ignore land.
+    "SST": OrderedDict(
+        [
+            # lambda sst: convert_units(rename(sst),target_units="degC")),
```
forsyth2, line 96: Why is this comment here?

e3sm_diags/derivations/utils.py
```diff
@@ -0,0 +1,309 @@
...
+    masked_var = var.where(cond=cond, drop=False)
+
+    return masked_var
+
+
+def adjust_prs_val_units(
+    prs: "FileAxis", prs_val: float, prs_val0: Optional[float]
+) -> float:
+    """Adjust the prs_val units based on the prs.id"""
+    # FIXME: Refactor this function to operate on xr.Dataset/xr.DataArray.
```
forsyth2, line 133: Is looks like this still has to be done?

e3sm_diags/derivations/utils.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def cosp_bin_sum(
+    cld: "FileVariable",
+    prs_low0: Optional[float],
+    prs_high0: Optional[float],
+    tau_low0: Optional[float],
+    tau_high0: Optional[float],
+):
+    # FIXME: Refactor this function to operate on xr.Dataset/xr.DataArray.
```
forsyth2, line 171: Is looks like this still has to be done?

e3sm_diags/derivations/utils.py
```diff
@@ -0,0 +1,309 @@
...
+        # cld_bin_sum.long_name = "{}: {} with {}".format(simulator, prs_lim, tau_lim)
+    except BaseException:
+        pass
+    return cld_bin_sum
+
+
+def determine_tau(
+    tau: "FileAxis", tau_low0: Optional[float], tau_high0: Optional[float]
+):
+    # FIXME: Refactor this function to operate on xr.Dataset/xr.DataArray.
```
forsyth2, line 227: Is looks like this still has to be done?

e3sm_diags/derivations/utils.py
```diff
@@ -0,0 +1,309 @@
...
+    else:
+        tau_low = tau_low0
+        tau_high = tau_high0
+        tau_lim = str(tau_low) + "< tau < " + str(tau_high)
+
+    return tau_high, tau_low, tau_lim
+
+
+def cosp_histogram_standardize(cld: "FileVariable"):
+    # TODO: Refactor this function to operate on xr.Dataset/xr.DataArray.
```
forsyth2, line 248: Is looks like this still has to be done?

forsyth2 (commented):
- [x] (1) Dataset class
- [x] (2) Computation, Regridding, and I/O Utilities
- [x] (3) Lat Lon Driver
- [x] (4) Lat Lon Plotter
- [x] (5) Derived Variables
- [x] (6) Misc.

I only reviewed the items mentioned in your PR description. Many of the functions are more complex than I can review line-by-line in a reasonable amount of time, but I tried to at least do a high-level read-through of each function. 

Thanks for working on this very large pull request!

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+        else:
+            raise ValueError(
+                f"The `type` ({self.type}) for this Dataset object is invalid."
+                "Valid options include 'ref' or 'test'."
+            )
+
+        # Set the `start_yr` and `end_yr` attrs based on the dataset type.
+        # Note, these attrs are different for the `area_mean_time_series`
+        # parameter.
+        if self.parameter.sets[0] in ["area_mean_time_series"]:
```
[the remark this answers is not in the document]
tomvothecoder: This logic was taken from the old `dataset.py` `Dataset` class:
https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/driver/utils/dataset.py#L241-L243

 I _think_ e3sm_diags was designed to run 1 set per parameter object, which means `parameter.sets` would contain one element. Example: 
https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/driver/default_diags/area_mean_time_series_model_vs_obs.cfg#L1-L6

It does seem fragile to have the conditional `if self.parameter.sets[0] in ["area_mean_time_series"]` because `CoreParameter.sets()` defaults to a list of all the sets with a specific order (below). I don't think we ever actually run e3sm_diags by passing a single parameter object with all of the sets like this though.

https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/parameter/core_parameter.py#L55-L76

@chengzhuzhang any thoughts?

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+            self.end_yr = self.parameter.test_end_yr  # type: ignore
+
+        # The derived variables defined in E3SM Diags. If the `CoreParameter`
+        # object contains additional user derived variables, they are added
+        # to `self.derived_vars`.
+        self.derived_vars_map = self._get_derived_vars_map()
+
+        # Whether the data is sub-monthly or not.
+        self.is_sub_monthly = False
+        if self.parameter.sets[0] in ["diurnal_cycle", "arm_diags"]:
```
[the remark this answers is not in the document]
tomvothecoder: Same comment about taking this logic from the old `Dataset` class and how we run e3sm_diags with one parameter object per set.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+                name = self.parameter.short_test_name
+            else:
+                name = self.parameter.test_name
+        elif self.type == "ref":
+            if self.parameter.short_ref_name:
+                name = self.parameter.short_ref_name
+            elif self.parameter.reference_name != "":
+                # parameter.ref_name is used to search though the reference
+                # data directories. parameter.reference_name is printed above
+                # Øref plots.
```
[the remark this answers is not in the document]
tomvothecoder: Accidental typo

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+                yrs_averaged = ""
+        elif self.is_time_series:
+            yrs_averaged = "{}-{}".format(self.start_yr, self.end_yr)
+
+        return yrs_averaged
+
+    def _get_derived_vars_map(self) -> DerivedVariablesMap:
+        """Get the defined derived variables.
+
+        If the user-defined derived variables is in the input parameters, append
```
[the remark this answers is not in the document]
tomvothecoder: Will fix

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+from e3sm_diags.parameter.core_parameter import CoreParameter
+
+logger = custom_logger(__name__)
+
+
+class Dataset:
+    def __init__(
+        self,
+        parameter: CoreParameter,
+        type: Literal["ref", "test"],
```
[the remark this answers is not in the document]
tomvothecoder: Fixed

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+            raise ValueError(
+                "The `season` argument is not a valid string. Options include: "
+                f"{CLIMO_FREQS}"
+            )
+
+        if self.is_climo:
+            ds = self._get_climo_dataset(season)
+        elif self.is_time_series:
+            ds = self.get_time_series_dataset(var)
+            ds[self.var] = climo(ds, self.var, season)
```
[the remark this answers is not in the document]
tomvothecoder: Updated `Dataset.is_time_series()` property to raise ValueError if data type can't be determined ([commit](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/c02d7015bbd1b211bdf3165d8412ace29ca23d5c)).

`Dataset.is_climo()` just checks if `Dataset.is_time_series()` is False.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+        str | None
+            The climatology filepath based on season, if it exists.
+        """
+        files_in_dir = sorted(os.listdir(root_path))
+
+        for file in files_in_dir:
+            if file.startswith(filename + "_" + season):
+                return os.path.join(root_path, file)
+
+        # The below is only ran on model data, because a shorter name is passed
```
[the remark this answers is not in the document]
tomvothecoder: Thanks, this was a legacy comment that I didn't catch.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+        # into this software. Won't work when use month name such as '01' as
```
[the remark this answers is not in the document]
tomvothecoder: Fixed

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+        Raises
+        ------
+        IOError
+            Multiple time series files found for the specified variable.
+        IOError
+            Multiple time series files found for the specified variable.
+        """
+        # The filename pattern for matching using regex.
+        # NOTE: Everything between '{var}_' and '.nc' in a time series file is
+        # always 13 characters.
```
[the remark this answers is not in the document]
tomvothecoder: This was pulled from the old `Dataset` class. I didn't refactor this logic.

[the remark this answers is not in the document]
tomvothecoder: Yup it works, will add a constant at the top of the file.

```python
LEN_TS_FILE_NAME = 13

var_key = "ts"
site = ["global"]
filename_pattern = var_key + "_" + sites[0] + rf"_.{LEN_TS_FILE_NAME}.nc"

print(filename_pattern)
# 'ts_global_.{13}.nc'
```

e3sm_diags/driver/utils/climo_xr.py
```diff
@@ -0,0 +1,156 @@
+"""This module stores climatology functions operating on Xarray objects.
+
+NOTE: Replaces `e3sm_diags.driver.utils.climo`.
+
+This file will eventually be refactored to use xCDAT's climatology API.
```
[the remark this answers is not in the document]
tomvothecoder: #726 is open for this

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+
+        # Get the available start and end years from the file name.
+        # Example: {var}_{start_yr}01_{end_yr}12.nc
+        var_start_year = int(filename.split("/")[-1].split("_")[-2][:4])
+        var_end_year = int(filename.split("/")[-1].split("_")[-1][:4])
+
+        if start_year < var_start_year:
+            raise ValueError(
+                "Invalid year range specified for test/reference time series data: "
+                f"start_year={start_year}>{var_start_year}=var_start_yr"
```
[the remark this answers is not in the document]
tomvothecoder: fixed

e3sm_diags/driver/utils/regrid.py
tomvothecoder: Yes, this will happen once all sets are refactored and no longer use `general.regrid_to_lower_res()`.

e3sm_diags/driver/utils/regrid.py
tomvothecoder: fixed

e3sm_diags/driver/utils/regrid.py
tomvothecoder: This was a question for @chengzhuzhang. The old logic assumes Z axis is always in mb units, but I wanted to make sure if this is correct.

e3sm_diags/driver/utils/regrid.py
tomvothecoder: I'll update the docstring to:

```python
The Z axis (e.g., 'lev') must either include hybrid-sigma levels or have 
pressure/isobaric coordinates to convert to pressure levels.
```

The logic below contains more context:
https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/driver/utils/regrid.py#L376-L389

e3sm_diags/driver/utils/regrid.py
tomvothecoder: fixed

e3sm_diags/metrics/metrics.py
tomvothecoder: Yes, this is a numpy docstring syntax to hyperlink to the "References" link if we ever host API docstrings.

https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/metrics/metrics.py#L150-L151

e3sm_diags/driver/utils/20230906-153411-issue-658-metrics-diff.xlsx
tomvothecoder: I will delete this file. I accidentally placed a copy of the Excel sheet generated by my metrics comparison script in the wrong directory.

e3sm_diags/driver/utils/io.py
tomvothecoder: No, this was untested code that was carried. I put a TODO statement here to make it clear it should be tested.

e3sm_diags/derivations/derivations.py
tomvothecoder: Not sure, this dictionary was pulled from `acme.py` into `derivations.py` without being refactored yet (should be handled in #716).

e3sm_diags/derivations/utils.py
tomvothecoder: Not yet, it needs to be refactored in #716.

e3sm_diags/derivations/utils.py
tomvothecoder: Not yet, it needs to be refactored in https://github.com/E3SM-Project/e3sm_diags/issues/716.

e3sm_diags/derivations/utils.py
tomvothecoder: Not yet, it needs to be refactored in https://github.com/E3SM-Project/e3sm_diags/issues/716.

e3sm_diags/driver/utils/io.py
tomvothecoder: This was a question for @chengzhuzhang. There is a section of logic for appending content to netCDF files with cdms2. Do we ever append data, or always write to a new file?

https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/driver/utils/general.py#L312-L333

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+                # All of the variables (list_of_vars) are in data_file.
+                # Return the corresponding dict.
+                return {tuple(var_list): target_variable_map[var_tuple]}
+
+        raise IOError(
+            f"The dataset file has no matching souce variables for {target_var}"
+        )
+
+    def _get_attr_from_climo(self, attr, season):
+        # TODO: Refactor this method.
```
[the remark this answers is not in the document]
tomvothecoder: No, this method is only used by the QBO set which will be another PR (#664).

I'll clarify the TODO statement to make it clear it is only used by QBO.

e3sm_diags/derivations/utils.py
tomvothecoder: I'll move all of these #716 related items over to this issue.

e3sm_diags/driver/utils/dataset_xr.py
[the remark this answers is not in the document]
chengzhuzhang: >I think e3sm_diags was designed to run 1 set per parameter object, which means parameter.sets would contain one element. '=
>
@tomvothecoder I think you are right. The sets are looped over looping over in `e3sm_diags_driver.run_diag()` and ran one at a time.

e3sm_diags/derivations/derivations.py
chengzhuzhang: "SST" is Sea Surface Temperature with land masked. I guess that's why the comment is there..

chengzhuzhang: OH, I think this refers to `# lambda sst ...` this can be cleaned up in later refactor.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -2,242 +2,427 @@
...
+            "min": diff_var.min().item(),
+            "max": diff_var.max().item(),
+            "mean": spatial_avg(ds_diff, var_key, serialize=True),
+        }
 
-    return parameter
+    return metrics_dict
+
+
+def _save_and_plot_metrics(
```
chengzhuzhang, line 373: how about to change to `_save_and_plot_data_metrics`?

e3sm_diags/driver/utils/io.py
chengzhuzhang: The cdms2 implementation saves one variable per file. I think aggregate test, ref (if available) and diff (if available) will be slightly cleaner? I don't have strong preference here.

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+    """
+    # TODO: Accept "esmf" as `tool` value for now because `CoreParameter`
+    # defines `self.regrid_tool="esmf"` by default and
+    # `e3sm_diags.driver.utils.general.regrid_to_lower_res()` accepts "esmf".
+    # Once this function is deprecated, we can remove "esmf" as an option here
+    # and update `CoreParameter.regrid_tool` to "xesmf"`.
+    if tool == "esmf":
+        tool = "xesmf"
+
+    lat_a = xc.get_dim_coords(ds_a[var_key], axis="Y")
```
chengzhuzhang, line 295: I'm wondering what error will be returned if user accidentally pass in unstructured datasets. I'm having this issue https://github.com/E3SM-Project/e3sm_diags/issues/165 in mind.

tomvothecoder: xCDAT uses CF attributes and `cf_xarray` to map to the coordinates via CF attributes.

**Each axis in a variable should only have one dimension coordinate assigned to it, which is determined by the CF metadata.** 

If a variable has multiple dim coordinates per axis, xCDAT will raise an error because there isn't a way to determine which one to return back. For example, this case will happen if a variable has two sets of lat coordinates (e.g., `lat1` and `lat2`) that conflict by sharing the same CF metadata (e.g., `axis="latitude"`, `standard_name="latitude"`).

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: Sounds good, I'll update the function name

e3sm_diags/derivations/derivations.py
tomvothecoder: Thanks for the clarification Jill

e3sm_diags/driver/utils/regrid.py
tomvothecoder: Moved to #734.

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: Actually, I'll update it to `_save_and_plot_metrics_dict()` because I call the dictionary variable `metrics_dict()` and the function `_create_metrics_dict()`.

e3sm_diags/driver/utils/regrid.py
chengzhuzhang: Yes the assumption is that Z axis, and in put units will all in units mb. Let's keep this TODO for now...

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+    hyam = _get_hybrid_sigma_level(ds, "hyam")
+    hybm = _get_hybrid_sigma_level(ds, "hybm")
+
+    if ps is None or hyam is None or hybm is None:
+        raise KeyError(
+            f"The dataset for '{var_key}' does not contain hybrid-sigma level 'ps', "
+            "'hyam' and/or 'hybm' to use for reconstructing to pressure data."
+        )
+
+    ps = _convert_units_to_mb(ps)
```
chengzhuzhang, line 501: at least for now all pressure related quantities are brought in/assumed to be in units `mb`

e3sm_diags/metrics/metrics.py
```diff
@@ -0,0 +1,233 @@
...
+import xskillscore as xs
+
+from e3sm_diags.logger import custom_logger
+
+logger = custom_logger(__name__)
+
+AXES = ["X", "Y"]
+
+
+def get_weights(ds: xr.Dataset):
```
chengzhuzhang, line 17: There is minor code inconsistency in computing weighted metrics, example, `std` computes weights in its own function, while `correlation`, `rmse` has weights passed in. I suggest to have weights computed in each metrics to be consistent. This means to compute weights one time more (?), but reduces the number of functions (`get_weights`) and arguments(`weights`)  being passed around..

e3sm_diags/parameter/core_parameter.py
```diff
@@ -96,7 +98,9 @@ def __init__(self):
         # Diagnostic plot settings
         # ------------------------
         self.main_title: str = ""
-        self.backend: str = "mpl"
+        # TODO: Remove `backend` because it is always e3sm_diags/plot/cartopy.
+        # This change cascades down to changes in `e3sm_diags.plot.plot`.
+        self.backend: str = "cartopy"
```
chengzhuzhang, line 103: Good call!

e3sm_diags/plot/utils.py
```diff
@@ -0,0 +1,459 @@
...
+
+            orig_fnm = os.path.join(
+                get_output_dir(parameter.current_set, parameter),
+                parameter.output_file,
+            )
+            fname = orig_fnm + ".%i." % idx + f
+            logger.info(f"Sub-plot saved in: {fname}")
+
+
+def _add_colormap(
```
chengzhuzhang, line 98: consider to change name to something like `_add_colormap_lat_lon`. Though it can be changed when other plotters being added, e.g. `zonal mean 2d` ...

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,921 @@
...
+        self.is_sub_monthly = False
+        if self.parameter.sets[0] in ["diurnal_cycle", "arm_diags"]:
+            self.is_sub_monthly = True
+
+    @property
+    def is_time_series(self):
+        is_ref_ts = self.parameter.ref_timeseries_input
+        is_test_ts = self.parameter.test_timeseries_input
+
+        if (self.data_type == "ref" and is_ref_ts) or (
```
chengzhuzhang, line 89: Maybe this `if` statement block can be simplified as:
```
if (self.data_type == "ref"):
    return is_ref_ts
else:
    return is_test_ts
```

chengzhuzhang (commented):
I'm finished my first pass of code review. Many nice changes in the refactor, more modular codes, clean doc strings and more unit tests! I left some minor comments and suggestions. Next I will try to run a few use cases with `lat_lon` to see if the new version breaks any thing..

e3sm_diags/driver/utils/regrid.py
chengzhuzhang: The metadata file of an example E3SM EAM native output looks like this:
```
netcdf v2.LR.historical_0101.eam.h0.2014-12 {
dimensions:
        ncol = 21600 ;
        time = UNLIMITED ; // (1 currently)
        nbnd = 2 ;
        chars = 8 ;
        lev = 72 ;
        ilev = 73 ;

variables:
        double lat(ncol) ;
                lat:long_name = "latitude" ;
                lat:units = "degrees_north" ;
        double lon(ncol) ;
                lon:long_name = "longitude" ;
                lon:units = "degrees_east" ;
          float ANRAIN(time, lev, ncol) ;
                ANRAIN:mdims = 1 ;
                ANRAIN:units = "m-3" ;
                ANRAIN:long_name = "Average rain number conc" ;
                ANRAIN:cell_methods = "time: mean" ;
```
I guess we can try it later to see if the error info is useful enough to tell the users an native datasets is passed in..

e3sm_diags/driver/lat_lon_driver.py
chengzhuzhang: in the plotter, both data and metrics are used, only only metrics. So maybe stick `data` in there?

e3sm_diags/driver/utils/regrid.py
tomvothecoder: Thanks for this example. xCDAT should be able to handle E3SM EAM native output fine because each of the variable's axis should only have 1 dimension coordinate with the appropriate CF attributes.

e3sm_diags/parameter/core_parameter.py
tomvothecoder: Yeah, the plan is to move all of the plots out of `plots/cartopy` one level down into `plots`.

I started that transition with `lat_lon_plot.py`, `lat_lon_land_plot.py` and `lat_lon_river_plot.py`.

e3sm_diags/driver/lat_lon_driver.py
```diff
@@ -2,242 +2,427 @@
...
-                for ilev in range(len(plev)):
-                    mv1 = mv1_p[
-                        ilev,
-                    ]
-                    mv2 = mv2_p[
-                        ilev,
-                    ]
+                for ilev, _ in enumerate(plev):
+                    ds_test_ilev = ds_test[ilev]
+                    ds_ref_ilev = ds_ref[ilev]
```
tomvothecoder, line 129: TODO: Fix this xarray subsetting syntax.

This syntax tries to get a variable based on `ilev`, rather than subset on `ilev`.
[`.isel`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.isel.html) instead.

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: Oh I think I see what you're saying. We're saving and plotting data _and_ metrics, and not appending "data" to "metrics". That makes sense to me, I'll make the change.

e3sm_diags/driver/utils/regrid.py
tomvothecoder: I think I pulled `_convert_units_to_mb` over from e3sm_to_cmip just in case units were in `mb`. 

I'll add a TODO to apply this function to the Z axis coordinates too.

e3sm_diags/driver/utils/dataset_xr.py
tomvothecoder: Thanks for the suggestion. I'll simplify the conditionals. I specifically added the `data_type` attr to make it easier to distinguish data types instead of using those other attrs.

e3sm_diags/driver/utils/io.py
tomvothecoder: I agree, it is cleaner to have a single netCDF that stores all variables. I'll update `_write_vars_to_netcdf()` to do this.

My question was if we ever appended data to existing files (old logic below)? I don't know if we ever do.
```python
 if Path(test_pth).is_file(): 
     cdms_arg = "a" 
```

e3sm_diags/metrics/metrics.py
chengzhuzhang: yes. The assumption is that both ds_a and ds_b should have the same grids/ area weights.

e3sm_diags/driver/utils/io.py
chengzhuzhang: I don't recall we ever appended data to existing files.

e3sm_diags/driver/utils/dataset_xr.py
tomvothecoder: I updated the conditional to:

```python
    @property
    def is_time_series(self):
        if self.parameter.ref_timeseries_input or self.parameter.test_timeseries_input:
            return True
        else:
            return False
```

since `self.data_type` must be initialized to `"ref"` or `"test"` here: 

https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/driver/utils/dataset_xr.py#L50-L59

e3sm_diags/plot/utils.py
tomvothecoder: `_add_color_map()` is the refactored version of the old `lat_lon_plot.plot_panel()`, which `aerosol_aeronet_plot.plot()` also uses.

I can update the `_add_color_map` docstring to say it is used by `lat_lon_plot.py` and will be used by `aersol_aeronet_plot.py` to make it clearer.

More context:
https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/plot/cartopy/deprecated_lat_lon_plot.py#L1-L7
https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/plot/cartopy/deprecated_lat_lon_plot.py#L72-L74
https://github.com/E3SM-Project/e3sm_diags/blob/c02d7015bbd1b211bdf3165d8412ace29ca23d5c/e3sm_diags/plot/cartopy/aerosol_aeronet_plot.py#L10

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,907 @@
...
+            if file.startswith(filename + "_" + season):
+                return os.path.join(root_path, file)
+
+        # The below is only ran on model data, because a shorter name is passed
+        # into this software. Won't work when use month name such as '01' as
+        # season.
+        if season in ["ANN", "DJF", "MAM", "JJA", "SON"]:
+            for file in files_in_dir:
+                if file.startswith(filename) and season in file:
+                    return os.path.join(root_path, file)
```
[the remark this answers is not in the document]
tomvothecoder: Updated `_get_climo_filepath()` docstring with this case.

e3sm_diags/driver/utils/dataset_xr.py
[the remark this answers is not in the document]
tomvothecoder: This can be moved over to another issue for possible refactoring.

e3sm_diags/driver/lat_lon_driver.py
tomvothecoder: Should be done with [`a19e9a4` (#677)](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/a19e9a4148b8ba18135104aae7ed80cc5a67e02d), just need to test.

e3sm_diags/metrics/metrics.py
tomvothecoder: Thanks, I updated these functions in [`a19e9a4` (#677)](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/a19e9a4148b8ba18135104aae7ed80cc5a67e02d#diff-3f01f708754161d2742b9affdeec02a41fef2e6f027c3175051262e06aedaad7). I also removed the `serialize` argument because we always serialize numpy arrays to a list of floats before writing them to`.json`.

e3sm_diags/driver/utils/io.py
tomvothecoder: Updated `_write_vars_to_netcdf()` to output test, ref, and diff variables in a single `.nc` file in [`e544788` (#677)](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/e544788cb305e5577446a5d1bf06f6e958082a17), which also includes unit tests.

e3sm_diags/driver/utils/dataset_xr.py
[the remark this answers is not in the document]
tomvothecoder: Actually, most of the sets use the method `Dataset.get_name_and_yrs()` which calls `_get_attr_from_climo()`.

I ended up refactoring these methods in [`9af60d0` (#677)](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/9af60d044c93865ab6ab79c455555c21cd38a1d2).

e3sm_diags/driver/utils/io.py
tomvothecoder: I added unit tests for this code in [`3363bd9` (#677)](https://github.com/E3SM-Project/e3sm_diags/pull/677/commits/3363bd9fc2573d194ad7b84c43ac03db195b205a)

e3sm_diags/driver/utils/regrid.py
```diff
@@ -0,0 +1,635 @@
...
+        return ds_a, ds_b_regrid
+    elif is_b_lower_res:
+        output_grid = ds_b.regridder.grid
+        ds_a_regrid = ds_a.regridder.horizontal(
+            var_key, output_grid, tool=tool, method=method
+        )
+
+        return ds_a_regrid, ds_b
+
+    # Both datasets have the same resolution, so return them without regridding.
```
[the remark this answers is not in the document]
tomvothecoder: In the old version, `regrid_to_lower_res()`, there is a comment that says:
 
https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/driver/utils/general.py#L246-L248

I interpreted this comment as also meaning that if the lat lengths are the same, the resolutions are the same. Is this comment incorrect or am I misinterpreting it?

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -0,0 +1,999 @@
...
+            The climatology filepath based on season, if it exists.
+        """
+        files_in_dir = sorted(os.listdir(root_path))
+
+        # If the filename is followed by _<SEASON>.
+        for file in files_in_dir:
+            if file.startswith(filename + "_" + season):
+                return os.path.join(root_path, file)
+
+        # For model only data, the <SEASON> string can by anywhere in the
```
chengzhuzhang, line 478: It seems like this code block can preserve the feature that when `ref_name` is given an empty string, the climo files can be still picked up based on the season string in the file? as mentioned [here](https://github.com/E3SM-Project/e3sm_diags/issues/736#issuecomment-1751458904).

e3sm_diags/driver/utils/regrid.py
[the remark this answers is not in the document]
chengzhuzhang: I think the assumption could be in general true that if nlat is smaller than the resolution would be higher. But I'm not sure if this is true all the time and if we should deal with exceptions. Let's make it an discussion item at the EZ meeting on Tues.

e3sm_diags/driver/utils/dataset_xr.py
tomvothecoder: Yeah I ported the code from here:
https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/driver/utils/dataset.py#L323-L326

but moved the `if season in ["ANN",...` outside of the loop to only check `season` once.

e3sm_diags/driver/utils/regrid.py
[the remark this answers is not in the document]
tomvothecoder: We decided to use the old logic for now to make sure both versions behave the same. We can consider making it more robust in this issue: https://github.com/E3SM-Project/e3sm_diags/issues/739

PR #694, opened 2023-06-17
Fix `numpy.int` --> `int`

xylar:
The aliases `numpy.float` and `numpy.int` are not supported in the latest version of `numpy`.


chengzhuzhang:
Thank you, Xylar!

PR #696, opened 2023-06-20
Simplify the conda env yml files

### LINKED ISSUE #695: Refine of 4 yaml files in `conda-env`
@tomvothecoder and @chengzhuzhang, I'd like to have a discussion about the purposes of the 4 different yaml files here.

### `ci.yml`
I think `ci.yml` probably needs more constraints on packages than it has in this PR, and certainly more than it had before this PR.  There's presumably a minimum `numpy`, `xarray`, etc. that `e3sm_diags` requires.

### `dev.yml`
I think `dev.yml` is too constrained.  I understand that the purpose is for everyone to be testing with the exact same versions of all the important packages but the reality of this ends up being that testing isn't happening with the newest available versions and thus problems (like the recent `numpy.int` and `numpy.float` issues) may not be noticed during developer testing and instead emerge in E3SM-Unified testing or after a release has happened.  I would instead suggest having `dev.yml` be the same as `ci.yml` with the additional dependencies needed for development.

### `docs.yml`
My only question here is whether the benefits of pinning to specific package versions (e.g. stability) outweigh the advantages of getting the latest versions instead.  This puts extra burden on developers to update the package versions over time.

### `prod.yml`
My question here is whether there is a clear advantage in folks using the same exact python and pip version to the patch level, and whether this is worth the effort from developers to keep it updated.  It seems easier to just put in the documentation:
```
conda env create -y -n e3sm_diags_env e3sm_diags
conda activate e3sm_diags_env
```

By the way there's an inconsistency between the name in the yaml file:
https://github.com/E3SM-Project/e3sm_diags/blob/main/conda-env/prod.yml#L2
and the docs:
https://e3sm-project.github.io/e3sm_diags/_build/html/main/install.html#a-latest-stable-release
And also the command in the docs should be:
```
conda env create -f prod.yml
```
without the `conda-env/` directory.

tomvothecoder:
- Closes #695 

This PR simplifies the conda env yml files by merging yml files together and loosening up constraints.

By no longer pinning exact dependencies, the conda environments will use the latest compatible dependencies. This ensures a higher chance of compatibility within E3SM Unified because we will be testing on similar dependency versions. 

The trade-off is we lose stability and deterministic environments between developers, which might make it a bit more difficult to reproduce bugs or behaviors. There are only a few developers in this repo so it's not that big of a deal to communicate environment setups and dependency versions.

Summary of Changes:

- Update `ci.yml` with minimum constraint for `numpy` and `xarray`
- Add doc dependencies for `ci.yml`
- Update `dev.yml` base section to match `ci.yml`
- Remove `prod.yml` and update docs with commands to get latest stable release of e3sm_diags
- Remove `docs.yml` and use `ci.yml` in GH Action builds
- Remove `prod.yml` section from `tbump.toml`


conda-env/dev.yml
```diff
@@ -6,25 +6,33 @@ channels:
...
+  # ==================
+  - scipy
+  - pytest
+  - pytest-cov
+  # Documentation
+  # =================
+  - sphinx
+  - sphinx_rtd_theme
+  - sphinx-multiversion
+  # Developer Tools
```
tomvothecoder, line 35: The QA tools such as `black` and `flake8` should still be pinned to exact versions because newer versions usually flag new warnings or errors. We should manually update these ourselves and fix the subsequent issues (if they are produced with newer versions).

conda-env/ci.yml
```diff
@@ -22,11 +22,16 @@ dependencies:
   - mache >=0.15.0
   - matplotlib-base
   - netcdf4
-  - numpy
-  - xarray
+  - numpy >=1.23.0
+  - xarray >=2023.02.0
```
tomvothecoder, line 26: I constrained `numpy >=1.23.0` because that one supports Python 3.11 (future-proofing).
I constrained `xarray >=2023.02.0` because that one drops Python 3.8 support.

conda-env/ci.yml
```diff
@@ -7,8 +7,8 @@ channels:
   - defaults
 dependencies:
   # Base
-  # ==================
-  - python>=3.8
+  # =================
+  - python >=3.9
```
tomvothecoder, line 11: Bumped minimum Python version to 3.9.

conda-env/dev.yml
```diff
@@ -6,25 +6,33 @@ channels:
...
+  # Testing
+  # ==================
+  - scipy
+  - pytest
+  - pytest-cov
+  # Documentation
+  # =================
+  - sphinx
+  - sphinx_rtd_theme
+  - sphinx-multiversion
```
tomvothecoder, line 34: This section replicates `ci.yml`

xylar:
Looks good to me.

tomvothecoder (commented):
My review comments

PR #703, opened 2023-07-08
Allow setting INSTALL_PATH using an env var

### LINKED ISSUE #684: Avoid hardwiring `INSTALL_PATH` so that installing the package is more flexible
More details TBD

mahf708:
This PR doesn't change any behavior; it only optionally adds a capability to allow setting INSTALL_PATH via env variable either at installation time or at runtime. 

The goal of this PR is twofold:

1. Allow setting INSTALL_PATH using an env variable
2. Point out the editable (`-e`) installation option for easier development

Currently, INSTALL_PATH works well for a conda env or an actual installation, but not editable installation. There's probably a much better way to organize the logic, but I opened the issue a while ago and never acted on it, so I decided to finally act on it. I will have this PR in draft mode for a bit first to get feedback from team members. I will also conduct my own tests locally to ensure it works well (I will update the text description with more information before marking it ready for review).

Fix #684

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -251,9 +251,17 @@ Instead, the developer will ``pip install .`` to build ``e3sm-diags`` with chang
 
         pip install .
 
-8. Check that tests pass: ``./tests/test.sh``. This takes about 4 minutes.
+8. If you need to make more than one edit, or have an "editable" installation (option ``-e``), use:
 
-9. Commit changes and make sure ``pre-commit`` checks pass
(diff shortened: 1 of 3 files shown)


xylar:
@mahf708, I'm not super expert in e3sm_diags but your changes look very reasonable to me.

xylar:
I had also tried a `-e` installation of `e3sm_diags` in the past and didn't have much luck. In other repos even with a `data_files` section in `setup()`, I've not had issues so it must be something specific to how it's used here.

Issue #652, opened 2023-02-16
Support short simulations 


chengzhuzhang:
Currently, E3SM Diags can only support analyzing seasonal and annual mean climatology. Based on discussions with @PeterCaldwell @crterai and @keziming. Adding support for short simulation will allow starting support EAMxx 1-40 day analysis and short EAM testing runs (i.g. one month). This is also requested by #348.

Things to consider for tackling this problem:

- The `seasons` parameter is used to find seasonal/annual mean climatology files if input are NCO processed climatology files, 
`seasons` is also used as target season if input files are NCO processed monthly time series and climo is computed on the fly.

- The supported `seasons`  in default viewer is hard-coded (i.e., https://github.com/E3SM-Project/e3sm_diags/blob/8ea5bad5f553a09b099df9464471b424eaa56b0c/e3sm_diags/viewer/default_viewer.py#L41),

- The `seasons` parameter is hard-coded in default configuration files (i.g., https://github.com/E3SM-Project/e3sm_diags/blob/main/e3sm_diags/driver/default_diags/lat_lon_land_model_vs_obs.cfg)

PR #712, opened 2023-07-27
Support monthly average and variable name convention for EAMxx

### LINKED ISSUE #348: Adding support for monthly mean climo in additional to seasonal/annual mean
As @milenaveneziani requested, it will be useful to look at monthly means plots.

chengzhuzhang:
Two use cases:
For longer than 1 year simulation:
Support both monthly average ['01', '02', ..., '12'] in viewer in addition to seasonal and annual mean.

For shorter than 1 year simulation:
Support visualization the averaged snapshot.

This PR is to support the first use case, it closes #348 and partially #652. 
This PR also add support for variable names from EAMxx to support its output.

An example run with regridded EAMxx output is shown [here](https://portal.nersc.gov/cfs/e3sm/zhang40/tests/lat_lon_eamxx_09/viewer/lat_lon/index.html). 
The run script for this run is linked [here](https://portal.nersc.gov/cfs/e3sm/zhang40/tests/lat_lon_eamxx_09/prov/run_lat_lon_eamxx.py).


chengzhuzhang:
Commit #[0c74dd9](https://github.com/E3SM-Project/e3sm_diags/pull/712/commits/0c74dd9d53fb903addc5f5f16581369799cb2fec) to enable 12 months means as default in core sets. I considered to add logic in season selector to support monthly means other than the original 5 (ANN, and 4 seasons), but it appears to adding these 12 month to each cfg files is less intrusive.

chengzhuzhang:
Note to myself: remember to update code [here](https://github.com/E3SM-Project/e3sm_diags/blob/a2d00eb114cf505a4ecf4dd429bb8710778cc087/e3sm_diags/driver/utils/dataset.py#L323) to support monthly data.

chengzhuzhang:
@kaizhangpnl Just for your information, this PR to support EAMxx regridded data in E3SM Diags is ready for review. I think you or someone from you team can be great early testers. I also included the result viewer in this thread for review. We can work together to refine this feature. 

On the other hand, @golaz has another PR in zppy to support processing native EAMxx output. We are planning to merge shortly. I will tag you again when the PR is ready to merge after some testing...

chengzhuzhang:
@tomvothecoder Tagging you for review. Changes from this PR might need to be carried to cdat migration branch at a later time..

kaizhangpnl:
> @kaizhangpnl Just for your information, this PR to support EAMxx regridded data in E3SM Diags is ready for review. I think you or someone from you team can be great early testers. I also included the result viewer in this thread for review. We can work together to refine this feature.
> 
> On the other hand, @golaz has another PR in [zppy](https://github.com/E3SM-Project/zppy/pull/424) to support processing native EAMxx output. We are planning to merge shortly. I will tag you again when the PR is ready to merge after some testing...

Great. I'd be happy to test it. Thanks for the useful information!

e3sm_diags/derivations/acme.py
```diff
@@ -654,9 +665,18 @@ def cosp_histogram_standardize(cld: "FileVariable"):
...
+                ("VapWaterPath",),  # EAMxx
+                lambda prw: convert_units(rename(prw), target_units="kg/m2"),
+            ),
+        ]
+    ),
+    "SOLIN": OrderedDict(
+        [
+            (("rsdt",), rename),
+            (("SW_flux_dn_at_model_top",), rename),  # EAMxx
         ]
```
tomvothecoder, line 678:
```suggestion
    "SOLIN": {
         ("rsdt",): rename
         ("SW_flux_dn_at_model_top",): rename,  # EAMxx
      }
```

I don't think you need need to use an `OrderedDict` in the nested derived variables dictionary. I just use a regular dictionary in my refactored version (`derivations.py`). The old Dataset class will probably be able to parse this the same way as the new Dataset class in the CDAT migration branch because order of the keys shouldn't matter.

https://github.com/E3SM-Project/e3sm_diags/blob/6b296b1fc4f8ddaea2a5efaa6639b95a31a009a8/e3sm_diags/derivations/derivations.py#L592-L604.

e3sm_diags/derivations/acme.py
```diff
@@ -913,29 +937,55 @@ def cosp_histogram_standardize(cld: "FileVariable"):
...
     "FSUTOA": OrderedDict([(("rsut",), rename)]),
     "FSUTOAC": OrderedDict([(("rsutcs",), rename)]),
     "FLNT": OrderedDict([(("FLNT",), rename)]),
-    "FLUTC": OrderedDict([(("rlutcs",), rename)]),
+    "FLUTC": OrderedDict(
+        [
+            (("rlutcs",), rename),
+            (("LW_clrsky_flux_up_at_model_top",), rename),  # EAMxx
+        ]
+    ),
```
tomvothecoder, line 954: Same comment about not needing to use an `OrderedDict`.

e3sm_diags/derivations/acme.py
tomvothecoder: I recognize you're updating existing elements that are already OrderedDict. Up to you if you want to use a regular dict for new or existing elements that you've modified. Just a suggestion since it is easier to read and cleaner, but I'm aware this file is going to be replaced anyways.

e3sm_diags/derivations/acme.py
```diff
@@ -1011,23 +1073,31 @@ def cosp_histogram_standardize(cld: "FileVariable"):
...
             (("hfls",), rename),
             (("QFLX",), lambda qflx: qflx_convert_to_lhflx_approxi(qflx)),
+            (("surface_upward_latent_heat_flux",), rename),  # EAMxx "s^-3 kg"
+        ]
+    ),
+    "SHFLX": OrderedDict(
+        [
+            (("hfss",), rename),
+            (("surf_sens_flux",), rename),  # EAMxx
         ]
```
tomvothecoder, line 1099: Same comment about not needing to use an `OrderedDict`.

e3sm_diags/derivations/acme.py
```diff
@@ -1454,7 +1524,12 @@ def cosp_histogram_standardize(cld: "FileVariable"):
     ),
     # Surface temperature: Degrees C
     # (Temperature of the surface (land/water) itself, not the air)
-    "TS": OrderedDict([(("ts",), rename)]),
+    "TS": OrderedDict(
+        [
+            (("ts",), rename),
+            (("surf_radiative_T",), rename),  # EAMxx
+        ]
+    ),
```
tomvothecoder, line 1532: Same comment about not needing to use an `OrderedDict`.

e3sm_diags/derivations/acme.py
```diff
@@ -1471,8 +1546,18 @@ def cosp_histogram_standardize(cld: "FileVariable"):
...
+            (("tauu",), lambda tauu: -tauu),
+            (("surf_mom_flux_U",), lambda tauu: -tauu),  # EAMxx
+        ]
+    ),
+    "TAUY": OrderedDict(
+        [
+            (("tauv",), lambda tauv: -tauv),
+            (("surf_mom_flux_V",), lambda tauv: -tauv),  # EAMxx
+        ]
+    ),
```
tomvothecoder, line 1560: Same comment about not needing to use an `OrderedDict`.

examples/run_v2_9_0_all_sets_E3SM_machines.py
```diff
@@ -69,7 +69,7 @@ def run_all_sets():
```
tomvothecoder, line 69: Do we need a new run script for the latest version of e3sm_diags?

tomvothecoder (approved):

e3sm_diags/derivations/acme.py
chengzhuzhang: Hi @tomvothecoder Thank you for reviewing. It's been a while since the ordered dictionary was designed. I think we made it ordered dictionary to support edge cases when some datasets include multiple variables in the key, and we wanted to loop over keys in order. I'm not sure if this is still needed given that lots of datasets are being updated over years.

examples/run_v2_9_0_all_sets_E3SM_machines.py
chengzhuzhang: I don't think so, we can just use the one run script but keep updating by adding new set.

e3sm_diags/derivations/acme.py
tomvothecoder: Maybe e3sm_diags development started around Python 3.5/3.6? I learned that the regular dict will maintain order with Python >=3.6, so we probably don't need to use `OrderedDict` anymore.

> Starting with CPython 3.6, dictionaries [return items in the order you inserted them](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation).
> Source - https://stackoverflow.com/a/47849121

> Since Python 3.7, all dictionaries are guaranteed to be ordered. The Python contributors determined that switching to making dict ordered would not have a negative performance impact. I don't know how the performance of OrderedDict compares to dict in Python >= 3.7, but I imagine they would be comparable since they are both ordered.
> 
> Note that there are still differences between the behaviour of OrderedDict and dict. See also: [Will OrderedDict become redundant in Python 3.7](https://stackoverflow.com/q/50872498/247696)?
> Source - https://stackoverflow.com/a/53535866

chengzhuzhang: Yeah, i think the development started around 3.5 and 3.6. In this case we should go with regular dictionary. Good call for the cdat migration!

tomvothecoder: Maybe `e3sm_diags` was developed around Python 3.5/3.6?

I learned that Python >=3.6 maintains order for regular dictionaries and that `OrderedDict` is probably not needed for our use-case anymore (get keys in order for matching in the dataset).

> Are dictionaries ordered in Python 3.6+?
> They are insertion ordered[1].
> 
> As of Python 3.6, for the CPython implementation of Python, dictionaries remember the order of items inserted. This is considered an implementation detail in Python 3.6; you need to use OrderedDict if you want insertion ordering that's guaranteed across other implementations of Python (and other ordered behavior[1]).
> 
> As of Python 3.7, this is a guaranteed language feature, not merely an implementation detail. [From a python-dev message by GvR](https://mail.python.org/pipermail/python-dev/2017-December/151283.html):
> Source: https://stackoverflow.com/a/39980744

tomvothecoder: No problem!

e3sm_diags/derivations/acme.py
```diff
@@ -705,6 +725,10 @@ def cosp_histogram_standardize(cld: "FileVariable"):
                 lambda fsntoa, fsntoac: swcf(fsntoa, fsntoac),
             ),
             (("rsut", "rsutcs"), lambda rsutcs, rsut: swcf(rsut, rsutcs)),
+            (
+                ("SW_flux_up_at_model_top", "SW_clrsky_flux_up_at_model_top"),
+                lambda rsutcs, rsut: swcf(rsut, rsutcs),
```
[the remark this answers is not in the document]
kaizhangpnl: The final "SWCF" plot looks correct, so I guess it's not an issue. 

https://portal.nersc.gov/cfs/e3sm/kzhang/e3sm_diag/lat_lon_eamxx_09/lat_lon/CERES-EBAF-TOA-v4.1/ceres_ebaf_toa_v4.1-SWCF-09-global.png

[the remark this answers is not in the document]
chengzhuzhang: Hi @kaizhangpnl thanks for looking into it! I was also supersized that the final plot looks reasonable, still investigating...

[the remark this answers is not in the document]
chengzhuzhang: Going through the code for derived variable and documentation for lamda expression. It does seem like the order of the parameters **after** `lambda` does not matter. 

In the derived variable dictionary, it seems like we can further simplify without using lamda expression, just to use the regular functions for most if not all cases. Tagging @tomvothecoder for confirmation..

kaizhangpnl (approved):
Tested it on Perlmutter and reproduced the plots.

chengzhuzhang:
> Tested it on Perlmutter and reproduced the plots.

Awesome! Thank you, Kai!

e3sm_diags/derivations/acme.py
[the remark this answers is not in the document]
tomvothecoder: > Going through the code for derived variable and documentation for lamda expression. It does seem like the order of the parameters **after** `lambda` does not matter.
> 
> In the derived variable dictionary, it seems like we can further simplify without using lamda expression, we can just to use the regular functions for most if not all cases. Tagging @tomvothecoder for confirmation..

That's correct, the lambda isn't needed because the required args will be passed directly to the function when it is called. The lambda is doing the same thing here.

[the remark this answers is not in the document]
chengzhuzhang: Thank you @tomvothecoder !

PR #713, opened 2023-08-01
wrap longitude 0-360 to fix iss711

### LINKED ISSUE #711: Difference lat-lon maps only show half for several obs datasets
In v2.9.0rc2, several lat-lon maps (bottom figure) only show half, including SST_HadISST, AOD_550 and OMI-MLS (@golaz's [new run](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.golaz/E3SM/fme/20230614.v2.LR.F2010/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_0022-0031/viewer/lat_lon/index.html) also shows this issue)
image

I saw the same issue when developing the GPCP v3.2 processing script. Rotating longitude from [-180,180) to [0, 360) resolved the issue. I suspect that a dependency library for visualization no longer auto consolidating  longitude with [-180,180) or [0, 360).

chengzhuzhang:
close #711

### DIFF
--- a/e3sm_diags/driver/utils/general.py
+++ b/e3sm_diags/driver/utils/general.py
@@ -243,6 +243,11 @@ def regrid_to_lower_res(mv1, mv2, regrid_tool, regrid_method):
     axes1 = mv1.getAxisList()
     axes2 = mv2.getAxisList()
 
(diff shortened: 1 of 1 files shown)


chengzhuzhang:
This PR is not needed, because the cause of issue #711  is that the cdms2 build was not compatible with newer version of esmf, and regridder was not imported correctly. Xylar patched `cdms2`, which solved the issue.

PR #722, opened 2023-08-28
Import shapely but not esmpy in main package

### LINKED ISSUE #720: concurrent.futures.process.BrokenProcessPool testing e3sm_unified rc12 on Perlmutter
`concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.` details documented here. https://github.com/E3SM-Project/zppy/issues/485#issuecomment-1686792065

Some troubleshooting suggested that the newly added `mp_partition` set (which is not included in zppy tests) triggered this problem. This issue was not seen in the e3sm_diags rc3 standalone environment, or in e3sm-unified rc12 pm-cpu login.

xylar:
This seems like it may fix #720.

### DIFF
--- a/e3sm_diags/__init__.py
+++ b/e3sm_diags/__init__.py
@@ -1,9 +1,10 @@
 import os
 import sys
 
-# import shapely before esmpy to prevent a segfault related to multiprocessing
-import shapely  # isort: skip
-import esmpy  # isort: skip
+# import shapely here (before any esmpy imports via cdms2) to prevent a
+# segfault related to multiprocessing.  Do not import esmpy here to prevent
+# issue with dask when using ESMF with system compilers.
+import shapely
 
 __version__ = "v2.9.0rc3"
 INSTALL_PATH = os.path.join(sys.prefix, "share/e3sm_diags/")


xylar:
@chengzhuzhang I believe I have a simple fix for e3sm_diags here.

This still fixes the issue in https://github.com/E3SM-Project/zppy/pull/475 because `shapely` gets imported first before any imports of `esmpy` (via `cdms2`).

It seems to fix the issue in https://github.com/E3SM-Project/e3sm_diags/issues/720 by not importing `esmpy` too early (maybe before `dask`).

My suggestion is that you make sure this branch works for you in some of your usual `e3sm_diags` standalone testing and then merge this PR and make another release candidate (rc4).

If you want to test in a configuration more similar to how it will run in E3SM-Unified, add these 2 commands after you load your conda environment but before you run `e3sm_diags` (as I did above):
``` bash
source /global/common/software/e3sm/anaconda_envs/spack/e3sm_unified_1_9_0rc13_pm-cpu_gnu_mpich/share/spack/setup-env.sh
spack env activate e3sm_unified_1_9_0rc13_pm-cpu_gnu_mpich
```

This will load the spack environment that E3SM-Unified uses.  It was sufficient to show the issues in https://github.com/E3SM-Project/e3sm_diags/issues/720 in my testing.

It would also be good to know if this solves the long run times on Chrysalis, but I think it will be easier to find that out after we do another deployment of E3SM-Unified.

chengzhuzhang:
@xylar It looks the timing for importing shapely and esmpy to avoid conflicts is important..Your initial tests look promising. I will do my regular standalone tests and plan to have a new rc today.

chengzhuzhang (approved):
@xylar I did my full e3sm_diags run test on Perlmutter, with e3sm_diags env and e3sm_diags env + spack env, then on Chrysalis. All 3 cases run successfully without errors. I have some difficulty get on Compy with the new double passwords requirement. But I think this PR should fix the problem and I'm ready for a new rc release!

chengzhuzhang:
Does activating `spack` env after activating e3sm_diags env  aggregate both environments? This is a new way that we can adopt to troubleshooting e3sm_diags with e3sm_unified.

forsyth2:
This is great, thanks @xylar  and @chengzhuzhang!

The only changes I have on `zppy` since the last releases are testing changes, so I don't think I need to do a new `zppy` RC. I will re-test `zppy` once we have a new E3SM Unified RC

PR #742, opened 2023-10-13
[Refactor]: Update `e3sm_diags_driver.run_diags` as `CoreParameter` method

tomvothecoder:
This PR updates the `run_diags()` function as a `CoreParameter` class method. This change is intended to make it clearer that `run_diags()` operates on a single `CoreParameter` object.

- Update VSCode workspace settings with extension and debugger configs

### DIFF
--- a/e3sm_diags/parameter/core_parameter.py
+++ b/e3sm_diags/parameter/core_parameter.py
@@ -241,7 +241,7 @@ class CoreParameter:
         for set_name in self.sets:
             self.current_set = set_name
             # FIXME: This is a shortcut to importing `run_diag`, but can break
-            # easily because the module driver must match the module name.
+            # easily because the module driver include the set name (static reference).
             # Instead, the import should be done more progammatically via
             # direct Python import.
             mod_str = "e3sm_diags.driver.{}_driver".format(set_name)
--- a/e3sm_diags/parameter/core_parameter.py
+++ b/e3sm_diags/parameter/core_parameter.py
@@ -1,19 +1,14 @@
 import copy
 import importlib
 import sys
-from typing import Dict, Generic, List, TypeVar
+from typing import Any, Dict, List
 
 from e3sm_diags.logger import custom_logger
 
 logger = custom_logger(__name__)
 
-T = TypeVar("T")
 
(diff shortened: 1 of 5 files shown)


e3sm_diags/parameter/core_parameter.py
```diff
@@ -213,3 +219,48 @@ def check_values(self):
...
+        List[Any]
+            The list of CoreParameter objects with results from the diagnostic run.
+            NOTE: `Self` type is not yet supported by mypy.
+        """
+        results = []
+
+        for set_name in self.sets:
+            self.current_set = set_name
+            # FIXME: This is a shortcut to importing `run_diag`, but can break
+            # easily because the module driver must match the module name.
```
tomvothecoder, line 244:
```suggestion
            # FIXME: This is a shortcut to importing `run_diag`, but can break
            # easily because the module driver include the set name (static reference).
```

e3sm_diags/parameter/core_parameter.py
```diff
@@ -213,3 +219,48 @@ def check_values(self):
         ):
             msg = "You need to define both the 'test_start_yr' and 'test_end_yr' parameter."
             raise RuntimeError(msg)
+
+    def _run_diag(self) -> List[Any]:
+        """Run the corresponding diagnostics for a set of parameters.
```
tomvothecoder, line 224:
```suggestion
        """Run the diagnostics for each set in the parameter.
```

e3sm_diags/parameter/core_parameter.py
```diff
@@ -213,3 +219,48 @@ def check_values(self):
...
+        List[Any]
+            The list of CoreParameter objects with results from the diagnostic run.
+            NOTE: `Self` type is not yet supported by mypy.
+        """
+        results = []
+
+        for set_name in self.sets:
+            self.current_set = set_name
+            # FIXME: This is a shortcut to importing `run_diag`, but can break
+            # easily because the module driver include the set name (static reference).
```
tomvothecoder, line 244:
```suggestion
            # easily because the module driver is statically imported via string.
```

PR #816, opened 2024-05-16
Update QBO amplitude calculation

chengzhuzhang:
Change the NFFT0, frequency, and interpolation of period calculations

## Description

@jjbenedict and @whannah1 identified an odd behavior that the QBO amplitude peak (bottom figure) gets weaker with longer years of simulation being analysed.  

@chengzhuzhang tested with multiple year lengths (1-10, 1-20, …1-100) and can reproduce this behavior. 
[QBO_diags_tests.pptx](https://github.com/E3SM-Project/e3sm_diags/files/15341563/QBO_diags_tests.pptx)

This PR is contributed by @justin-richling and Jack Chen (cchen@ucar.edu) for coming up with a fix.  

- Closes #<ISSUE_NUMBER_HERE>

## Checklist

If applicable:

### DIFF
--- a/e3sm_diags/driver/qbo_driver.py
+++ b/e3sm_diags/driver/qbo_driver.py
@@ -129,12 +129,12 @@ def get_psd_from_deseason(xraw, period_new):
     # Calculate the period as a function of frequency
     period0 = 1 / sampling_frequency
     L0 = len(xraw)
-    NFFT0 = 2 ** ceil_log2(L0)
+    NFFT0 = L0
 
     # Apply fft on x_deseasoned with n = NFFT
     x0 = scipy.fftpack.fft(x_deseasoned, n=NFFT0) / L0
     # Frequency (cycles/month). Frequency will be increasing.
-    frequency0 = sampling_frequency * np.arange(0, (NFFT0 / 2 + 1)) / NFFT0
+    frequency0 = (sampling_frequency/2)*np.linspace(0, 1, int(NFFT0 / 2)+1)
     # Period (months/cycle). Calculate as a function of frequency. Thus, period will be decreasing.
     period0 = 1 / frequency0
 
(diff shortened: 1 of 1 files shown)


chengzhuzhang:
@justin-richling Thanks for the PR! I still won't be able to figure out why it is needed to add the `default=str` argument when json.dump (`json.dump(json_dict, outfile, default=str)`). But it seems to be a fine workaround for saving the files in json format. 

I ran a test with the new algorithm and compared to the original one, the power spectrum density changed a lot, and I'm not sure if it is the intended behavior. Could you provide an explanation of the code change here? Thank you!

Before:

https://portal.nersc.gov/cfs/e3sm/chengzhu/tutorial2024/e3sm_diags_extended/qbo/QBO-ERA-Interim/qbo_diags.png

After:
image

justin-richling:
Sure thing, @chengzhuzhang!

From Jack:

Fast Fourier transform (FFT) is used for the QBO spectra calculator in which power spectra in frequency space is carried out in a discrete form. 

The original code used an interpolation approach to obtain power spectra in the period space (inverse of frequency).

However, this approach could not produce consistent results under simulations of different durations.
For a longer simulation, more frequencies will be resolved through FFT which could lower the peak amplitude in the frequency space.

The revised code uses a bin approach which sums all the power within each period bin (2 months).
Consequently, this approach produces robust power spectra.

jjbenedict:
@justin-richling -- so it looks like in this 15-year demonstration case the higher frequency activity is better resolved and the lower frequency activity is poorly resolved.  I assume amplitude line in the bottom panel would become smoother if a longer time window were used, correct?

justin-richling:
@jjbenedict Correct, for something like a typical 30-year run (ERA Interim), the result will be smoother.

chengzhuzhang:
Thank you for the explanation @justin-richling!
My understanding is that the original one does provide smoother power spectrum, but the amplitude of the peak can get weaker when more frequency is resolved given longer time series. The new one could be difficult to use for simulations shorter than i.e. 30 years. @jjbenedict Could you provide some recommendation how we should move forward?

jjbenedict:
@chengzhuzhang I talked with @justin-richling and Jack earlier this week.  Justin is actively working with Walter Hannah to implement the wavelet analysis (Walter had earlier written a python script to compute the wavelets).  We're thinking it might be good to include in E3SM Diags both the traditional Fourier method (as implemented now, and appropriate for time windows of 30 yr or longer) -and- the wavelet method (which may be better for shorter time windows).  We'll have to do some tests with the wavelet method to see how effective it is at shorter time windows, but the expectation is that is will be smoother than Jack's existing Fourier method.  Maintaining the Fourier method would also allow greater consistency with QBO plots from previous E3SM versions.  We realize the urgency of implementing the wavelet analysis, given that v3 simulations are now being run... we'll do our best to get this set up and tested as quickly as possible.

FYI -- I will be on vacation July 6-22 (though interrupted by a meeting July 16-17, when I will be checking email).

chengzhuzhang:
@jjbenedict thank you for the updates, and the heads-up! It seems a good idea to include both methods, and we can discuss on the implementation details after getting testing results from the wavelet analysis.

minxu74:
Besides the smoothness, I wonder why the phase were different among the old and revised ones.  The old result showed the phase shifted simulated by E3SM, but the new one looked perfectly good.

PR #832, opened 2024-07-31
CDAT Migration Phase 2: Refactor `enso_diags` set

### LINKED ISSUE #663: CDAT Migration Phase 2: Refactor `enso_diags` set
Just noting that @forsyth2 previously worked on this set.

tomvothecoder:
## Description

- Closes #663 

## TODO

8/14/24
- [x] Refactor map driver
- [x] Add optional kwargs `plot_kwargs` and `viewer_descr` to `_save_data_metrics_and_plots()` to support custom enso diags configs
- [x] Refactor map plotter 
  - [x] Get it working for a variable with saved outputs
  - [x] Consider refactoring portions of the code to reuse general plotter functions
- [x] Refactor scatter driver 
- [x] Refactor scatter plotter 

8/20/24
- [x] Regression testing with all variables
- [ ] ~~"ccb" not being considered for regional subsetting "20S20N" --> 2 extra points causes diffs in floating points and plots.~~ won't fix based on Jill's comment [below](https://github.com/E3SM-Project/e3sm_diags/pull/832#issuecomment-2299788687)

8/21/24
- [x] Fix unit test
- [x] Fix integration test 
  - [x] `OSError: No files found for target variable SST or derived variables ([('sst',), ('TS', 'OCNFRAC'), ('SST',)]) in tests/integration/integration_test_data.` -- due to incorrect exception
  - [x] `AttributeError: Either "parameter.short_ref_name", "parameter.reference_name", or "parameter.ref_name" must be set to get the name and years attribute for reference datasets.` -- need to preserve the `else` statement from the previous logic that sets the test and ref names to `parameter.test_name`/ `parameter.ref_name` regardless if either is set or not (empty string).

## Checklist

If applicable:


tomvothecoder:
@chengzhuzhang I'm running into the same issue found in #664 where the `"ccb"` flag adds an additional point to the start and end of the `lat` coordinates, resulting in mismatching output arrays. Once I remove the two additional points, the variables align between `main` and this branch.

I'm trying to write a adapt the function I wrote to handle the `"ccb"` flag for the lat dimension. It is not clear how cdms2 is calculating the additional start and end points (time delta between bounds is not used unlike `"ccb"` with time coordinates). I'll need to investigate the code in cdms2 unfortunately. 

Any thoughts?

tomvothecoder:
@chengzhuzhang What is the purpose of using the `"ccb"` flag for non-submonthly data in the CDAT code? I still don't understand the reasoning.

chengzhuzhang:
@tomvothecoder I think it is just a convenience choice to use `ccb` in this set. I don't think it is important to exactly matching the number of lat lon points in refactored code to the `main`. 

To support flag like `ccb` can be investigated in `xcdat`, but here I don't see it is necessary to have an exact match.

tomvothecoder:
> I think it is just a convenience choice to use `ccb` in this set.

@chengzhuzhang What I mean is: what is the purpose of the CDAT slice flag? I'm not clear on why we need the extra coordinates points when opening time series datasets using `ccb`.

> I don't think it is important to exactly matching the number of lat lon points in refactored code to the `main`.

Got it

chengzhuzhang:
@tomvothecoder could you point me to the original codes where `ccb` are used, is it used for both time and lat/lon dimensions?

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -1238,66 +1237,118 @@ def _get_time_slice(self, ds: xr.Dataset, filename: str) -> slice:
...
-        time_bnds_key = time_coords.attrs["bounds"]
-
-        # Extract the sub-dataset for all data at the last time coordinate and
-        # get the delta between time coordinates using the difference between
-        # bounds values.
-        ds_last_time = ds.isel({time_dim: -1})
-        time_bnds = ds_last_time[time_bnds_key]
-        time_delta = time_bnds[-1] - time_bnds[0]
+        actual_day = time_coords[0].dt.day.item()
+        actual_month = time_coords[0].dt.month.item()
```
tomvothecoder, line 1294:
```suggestion
		first_time = time_coords[0]
        actual_day = first_time.dt.day.item()
        actual_month = first_time.dt.month.item()
        actual_year = first_time.dt.year.item()
```

tomvothecoder:
@chengzhuzhang This PR should be good to go after the build passes. Let me know if you have any comments or questions based on my testing results.

## Regression testing results (netCDF) -- should be good to go
- Regression-coefficient -- within rtol once the extra two lat coordinates are removed from the CDAT files (added via "ccb" slice flag)
- Feedback -- rtol is large because we are comparing anomalies. I compared the min, max, mean, and std and they are all pretty close for one file. I think if I compared the spatial average before calculating anomalies instead then the rtol would be met. _Note: CDAT code on `main` doesn't save these variables out to netCDF so I made a temporary manual change for the comparison._ 

## Regression test (plots) -- any diffs are not discernible via eye-check

`main`: https://portal.nersc.gov/project/e3sm/cdat-migration-fy24/663-enso-diags/viewer/enso_diags/index.html
 dev branch: https://portal.nersc.gov/project/e3sm/cdat-migration-fy24/main/viewer/enso_diags/index.html

- Regression-coefficient -- mostly the same, except there is a small white bar for the xCDAT plot (I think due to the missing 2 lat points). Example diff for SHFLX on [`main`](https://portal.nersc.gov/project/e3sm/cdat-migration-fy24/main/viewer/enso_diags/map/shflx-response/plot.html) vs. [dev branch](https://portal.nersc.gov/project/e3sm/cdat-migration-fy24/663-enso-diags/viewer/enso_diags/map/shflx-response/plot.html)
regression-coefficient-shflx-over-nino34
-  Feedback -- mostly the same, except data points and lines are moved over just a very small amount (I think due to the missing 2 lat points)

tomvothecoder:
> @tomvothecoder could you point me to the original codes where `ccb` are used, is it used for both time and lat/lon dimensions?

`ccb` is used in `default_regions.py`. Example:

https://github.com/E3SM-Project/e3sm_diags/blob/b77d681280e6825eb740af89705ac0888820a0ab/e3sm_diags/derivations/default_regions.py#L132

chengzhuzhang:
Thanks, @tomvothecoder . Okay, it is for the NINO index. I think as long as the refactored code is in line with the [xCDAT example](url) Steve Po-Chedley provided (https://xcdat.readthedocs.io/en/latest/examples/spatial-average.html#4.-Nino-3.4-Region). I think we should be good.

chengzhuzhang:
@tomvothecoder I think there is a problem with the metrics (max, min, mean) plotted in wrong order in the refactored code.

tomvothecoder:
> @tomvothecoder I think there is a problem with the metrics (max, min, mean) printed in wrong order in the refactored code. Also I can't seem to find other plots, only Regression-coefficient plot for PRECT is available in the dev branch results... image

I ran the run script again. All of them are there now: https://portal.nersc.gov/project/e3sm/cdat-migration-fy24/663-enso-diags/viewer/enso_diags/index.html

Thanks for catching the issue with the max and min values needing to be swapped. I also noticed RMSE and CORR aren't printed. Those should be easy fixes and will be addressed via #838.

PR #851, opened 2024-09-24
Fixing African easterly wave density plots in TC analysis

chengzhuzhang:
## Description
Karthik (@karthik-balaguru) noted the odd Y shapes in easterly wave density plots in recent TC analysis results (issue 1). And @paullric also noticed that the ERA5 results are also off, which always have a base value of ~5 (issue 2). 

image

- Closes #<ISSUE_NUMBER_HERE>

## Checklist

If applicable:

### DIFF
--- a/e3sm_diags/driver/tc_analysis_driver.py
+++ b/e3sm_diags/driver/tc_analysis_driver.py
@@ -160,9 +160,14 @@ def generate_tc_metrics_from_te_stitch_file(te_stitch_file: str) -> Dict[str, An
     """
     logger.info("\nGenerating TC Metrics from TE Stitch Files")
     logger.info("============================================")
+    if not os.path.exists(te_stitch_file):
+        raise FileNotFoundError(f"The file {te_stitch_file} does not exist.")
 
     with open(te_stitch_file) as f:
         lines_orig = f.readlines()
+
+    if not lines_orig:
+        raise ValueError(f"The file {te_stitch_file} is empty.")
 
     line_ind = []
     data_start_year = int(te_stitch_file.split(".")[-2].split("_")[-2])
--- a/e3sm_diags/driver/tc_analysis_driver.py
+++ b/e3sm_diags/driver/tc_analysis_driver.py
@@ -165,6 +165,7 @@ def generate_tc_metrics_from_te_stitch_file(te_stitch_file: str) -> Dict[str, An
         lines_orig = f.readlines()
 
     line_ind = []
+    data_start_year = int(te_stitch_file.split(".")[-2].split("_")[-2])
     data_end_year = int(te_stitch_file.split(".")[-2].split("_")[-1])
     for i in range(0, np.size(lines_orig)):
         if lines_orig[i][0] == "s":
@@ -181,6 +182,13 @@ def generate_tc_metrics_from_te_stitch_file(te_stitch_file: str) -> Dict[str, An
     num_storms, max_len = _calc_num_storms_and_max_len(lines)
     # Parse variables from TE stitch file
     te_stitch_vars = _get_vars_from_te_stitch(lines, max_len, num_storms)
+    # Add year info
+    te_stitch_vars["year_start"] = data_start_year
+    te_stitch_vars["year_end"] = data_end_year
+    te_stitch_vars["num_years"] = data_end_year - data_start_year + 1
+    logger.info(
+        f"TE Start Year: {te_stitch_vars['year_start']}, TE End Year: {te_stitch_vars['year_end']}, Total Years: {te_stitch_vars['num_years']}"
+    )
 
     # Use E3SM land-sea mask
     mask_path = os.path.join(e3sm_diags.INSTALL_PATH, "acme_ne30_ocean_land_mask.nc")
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
Not sure the root cause of issue 2, but I think plotting libraries changes caused the second issue over versions. By reading in files without subsetting the region, it fixed the issue.
https://github.com/E3SM-Project/e3sm_diags/pull/851/commits/3fefb76575115d74106fb019c53dc50e9ac51126
image

Issue 1 remains, I will keep investigating.

chengzhuzhang:
Comparing with apply `tempest-extremes` (TE) on native ne120pg2 vs data regrided to 720x1440. The Y shapes only appear with native grid data.  Below are the AEW density plots. 
awe_ne120pg2
awe_720x1440

The AEW density plots uses three TE functions:
`VariableProcessor` to calculate relative vorticity
`DetectNodes` and `StitchNodes` to detect and stitch candidates
Finally `HistogramNodes` to generate histogram for the detection.
Detailed steps can be found [here](https://github.com/E3SM-Project/e3sm_diags/blob/2b303dfdf1a83b115a8f372a81bf2f6b3bfbbc68/model_data_preprocess/postprocessing_E3SM_data_for_TC_analysis.sh#L80-L94)

@paullric I'm not sure if TE is applied properly in this case, any advice for further trouble shooting is appreciated!

chengzhuzhang:
Also to add that:
TC count increases if the storms are tracked on the regridded output. ~73 (regridded data on 720x1440) vs ~67 (native ne120) for one year of data.

chengzhuzhang:
> Also to add that: TC count increases if the storms are tracked on the regridded output. ~73 (regridded data on 720x1440) vs ~67 (native ne120) for one year of data.

Following suggestion from @wlin7, when regridding, with a bilinear mapping file (to replace the aave one I used), the TC count (70) get closer to the results tracked with native ne120 data.

chengzhuzhang:
This PR will also address: empty input TC files.
from comment: https://github.com/E3SM-Project/zppy/pull/633#issuecomment-2423038746

chengzhuzhang:
> This PR will also address: empty input TC files. from comment: [E3SM-Project/zppy#633 (comment)](https://github.com/E3SM-Project/zppy/pull/633#issuecomment-2423038746)

This is addressed by https://github.com/E3SM-Project/e3sm_diags/pull/866/commits/0cb36e971c32db4caa3002bc184fff2e1df895db  by @tomvothecoder in cdat-migration-branch. I will copy the code over here, so that current `main` branch also has this fix.

chengzhuzhang:
@tomvothecoder this PR is ready to review, it partially addresses the plotting problem (Y shapes remains); also cross year TCs are removed; I copied over the error capture for cyclones stitches files as well.

e3sm_diags/driver/tc_analysis_driver.py
```diff
@@ -163,13 +160,40 @@ def generate_tc_metrics_from_te_stitch_file(te_stitch_file: str) -> Dict[str, An
...
+    lines = lines_orig[0:end_ind]
 
     # Calculate number of storms and max length
     num_storms, max_len = _calc_num_storms_and_max_len(lines)
     # Parse variables from TE stitch file
     te_stitch_vars = _get_vars_from_te_stitch(lines, max_len, num_storms)
+    # Add year info
+    te_stitch_vars["year_start"] = data_start_year
+    te_stitch_vars["year_end"] = data_end_year
+    te_stitch_vars["num_years"] = data_end_year - data_start_year + 1
```
tomvothecoder, line 193: If I understand correctly, you moved the year info out of `_get_vars_from_te_stitch()` because they are now different after removing the excessive time points that cross year bounds. Is this right?

e3sm_diags/driver/tc_analysis_driver.py
```diff
@@ -163,13 +160,40 @@ def generate_tc_metrics_from_te_stitch_file(te_stitch_file: str) -> Dict[str, An
...
+    for i in range(0, np.size(lines_orig)):
+        if lines_orig[i][0] == "s":
+            year = int(lines_orig[i].split("\t")[2])
+
+            if year <= data_end_year:
+                line_ind.append(i)
+
+    # Remove excessive time points cross year bounds from 6 hourly data
+    end_ind = line_ind[-1]
+    lines = lines_orig[0:end_ind]
```
tomvothecoder, line 184: Minor suggestion, I would extract this block of logic into a private function called `_remove_time_crossing_bounds()` or something to make the main `run_diags()` cleaner. 

However, this can be done in the refactored codebase too.

tomvothecoder (approved):
LGTM with some comments.

e3sm_diags/driver/tc_analysis_driver.py
chengzhuzhang: The start and end years were previously based on data from te_stitches files. It worked fine earlier because in the simulated data, there are always TCs each year. It became problematic when testing v3 datasets, years with no TCs get skipped. The new code now use start and end included in file name, which represent the actually year range being assessed.

e3sm_diags/driver/tc_analysis_driver.py
chengzhuzhang: Thank you for the suggestion. I considered to have a function instead as well.  I think we can update the refactored code base when bringing more fix and enhancement to tc set.

PR #952, opened 2025-03-12
Initial prototype of dask distributed to replace dask.bag

### LINKED ISSUE #948: [Feature]: Consider replacing Python multiprocessing with Dask distributed for performance
### Is your feature request related to a problem?

E3SM Diags performs parallel diagnostic runs on a per variable level. It utilizes a function called `_run_with_dask()` which passes the list of diagnostic runs to `dask.bag`. `dask.bag` is essentially a wrapper of Python multiprocessing. 

https://github.com/E3SM-Project/e3sm_diags/blob/e2fcc3abc5a984661b978d3082db1e42ab5559a8/e3sm_diags/e3sm_diags_driver.py#L334-L376

Currently, e3sm_diags loads datasets into memory via `.load()`. This means we can't chunk datasets and perform any downstream operations in parallel (e.g., spatial averaging).

---

If we can chunk datasets and perform operations on datasets in parallel, this might speed up `e3sm_diags` performance. 

### Describe the solution you'd like

I started experimenting with replacing `dask.bag` with dask's distributed cluster (related [discussion](https://github.com/E3SM-Project/e3sm_diags/issues/933#issuecomment-2649474119)).

## Asking ChatGPT: Should I replace dask.bag with a dask distributed scheduler?

> 
> Yes, replacing `dask.bag` with **Dask Distributed** is a good idea because it provides better **process isolation**, **fault tolerance**, and **MPI compatibility**. Here’s why:
> 
> ---
> 
> ### **Why Use Dask Distributed Instead of `dask.bag` with Multiprocessing?**
> #### **1. Avoids `multiprocessing`-related Crashes**
> - `dask.bag` with `multiprocessing` relies on `concurrent.futures.ProcessPoolExecutor`, which can lead to **MPI conflicts** and the `BrokenProcessPool` error.
> - Dask Distributed runs **each worker in its own separate process**, avoiding `fork()` issues.
> 
> #### **2. Better Handling of MPI**
> - Dask Distributed **isolates workers properly**, preventing unwanted MPI reinitialization.
> - Unlike Python’s `multiprocessing`, Dask Distributed doesn’t inherit an already-initialized MPI environment across processes.
> 
> #### **3. More Robust and Scalable**
> - Supports **dynamic task scheduling**, while `dask.bag` is more static.
> - Handles **worker failures gracefully** (e.g., if a process crashes, the scheduler can restart it).
> 
> #### **4. More Control Over Worker Processes**
> - With Dask Distributed, you can:
>   - Specify how many workers to use.
>   - Run a scheduler on a different machine.
>   - Monitor task execution with **Dask’s dashboard**.
> 
> ---
> 
> ### **How to Replace `dask.bag` with Dask Distributed**
> #### **Step 1: Start a Dask Distributed Client**
> Modify your script to use `dask.distributed` instead of `dask.bag` with multiprocessing:
> ```python
> from dask.distributed import Client
> 
> if __name__ == "__main__":
>     client = Client()  # Automatically sets up a distributed scheduler
>     print(client)  # Shows connection info
> 
>     # Run your Dask computation here...
> ```
> - This automatically **creates a local cluster** with multiple workers.
> 
> #### **Step 2: Convert `dask.bag` to Use `dask.delayed` or Dask DataFrame**
> Instead of:
> ```python
> import dask.bag as db
> b = db.from_sequence(data).map(my_function)
> result = b.compute()
> ```
> Use:
> ```python
> from dask import delayed
> 
> tasks = [delayed(my_function)(item) for item in data]
> result = dask.compute(*tasks)
> ```
> - This approach **distributes the workload across the Dask cluster**.
> 
> #### **Step 3: Use `client.submit` for Parallel Execution**
> For better control, use `client.submit`:
> ```python
> futures = client.map(my_function, data)
> results = client.gather(futures)  # Collect results
> ```
> - This avoids the limitations of `dask.bag`.
> 
> ---
> 
> ### **Summary: Should You Replace `dask.bag`?**
> ✅ **Yes**, because:
> - Dask Distributed is **more stable with MPI**.
> - It provides **better process isolation**.
> - It supports **dynamic, fault-tolerant task execution**.
> 
> Would you like a more detailed example with your specific use case? 🚀

### Describe alternatives you've considered

_No response_

### Additional context

There are some MPI issues in #933 to be aware of.

tomvothecoder:
## Description

- Closes #948 

## Checklist

If applicable:

### DIFF
--- a/auxiliary_tools/949_dask_distributed/v2_run.cfg
+++ b/auxiliary_tools/949_dask_distributed/v2_run.cfg
@@ -1,0 +1,385 @@
+[#]
+sets = ["lat_lon"]
+case_id = "ERA5"
+variables = ["TREFHT"]
+regions = ["land"]
+ref_name = "ERA5"
+reference_name = "ERA5 Reanalysis"
+seasons=["ANN"]
+contour_levels = [-35, -30, -25, -20, -15, -10, -5, 0, 5, 10, 15, 20, 25, 30, 35, 40]
+diff_levels = [-12, -8, -4, -2, -1, -0.5, 0.5, 1, 2, 4, 8, 12]
+regrid_method = "bilinear"
+
+# [#]
+# sets = ["lat_lon"]
+# case_id = "ERA5"
+# variables = ["QREFHT"]
+# ref_name = "ERA5_ext"
+# reference_name = "ERA5 Reanalysis"
+# seasons=["ANN"]
+# seasons = ["ANN"]
+# contour_levels = [0.2, 0.5, 1, 2.5, 5, 7.5, 10, 12.5, 15, 17.5]
+# diff_levels = [-5, -4, -3, -2, -1, -0.25, 0.25, 1, 2, 3, 4, 5]
+
+# [#]
+# sets = ["lat_lon"]
+# case_id = "GPCP_v3.2"
+# variables = ["PRECT"]
+# ref_name = "GPCP_v3.2"
+# reference_name = "GPCP_v3.2"
+# seasons=["ANN"]
+# regions = ["global"]
+# test_colormap = "WhiteBlueGreenYellowRed.rgb"
+# reference_colormap = "WhiteBlueGreenYellowRed.rgb"
+# diff_colormap = "BrBG"
+# contour_levels = [0.5, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 13, 14, 15, 16]
+# diff_levels = [-5, -4, -3, -2, -1, -0.5, 0.5, 1, 2, 3, 4, 5]
+
+# [#]
+# sets = ["lat_lon"]
+# case_id = "SST_CL_HadISST"
+# variables = ["SST"]
+# ref_name = "HadISST_CL"
+# reference_name = "HadISST (Climatology)"
+# seasons=["ANN"]
+# contour_levels = [-1, 0, 1, 3, 6, 9, 12, 15, 18, 20, 22, 24, 26, 28, 29]
+# diff_levels = [-5, -4, -3, -2, -1, -0.5, -0.2, 0.2, 0.5, 1, 2, 3, 4, 5]
+# regrid_method = "bilinear"
+
+# [#]
+# sets = ["lat_lon"]
+# case_id = "SST_PD_HadISST"
+# variables = ["SST"]
+# ref_name = "HadISST_PD"
+# reference_name = "HadISST (Present Day)"
+# seasons=["ANN"]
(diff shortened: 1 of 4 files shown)


tomvothecoder:
I am running into an issue where utilizing xESMF/ESMF breaks when using a Dask Distributed scheduler and more than 1 variable is being regridded. This also might be related to #988. 

> **ESMF + xESMF failure when using Dask Distributed**
> 
> When using `xESMF` with a Dask Distributed scheduler, I encountered the following error:
> 
> ```
> ESMCI::VM::getCurrent() Internal error: Bad condition  - Could not determine current VM
> ESMF_VMGetCurrent() Internal error: Bad condition
> ESMF_GridCreateNoPeriDimR Internal error: Bad condition
> ...
> ```
> 
> This is triggered when **`xESMF.Regridder` is initialized inside a Dask worker**, which leads to race conditions in the ESMF Fortran backend (e.g., `ESMF_GridCreateDistgridReg`, `VMGetCurrent`, etc). ESMF is not thread- or process-safe during grid/regridder setup.
> 
> ### ✅ **Workaround**
> 
> Initialize all `Regridder` objects **serially** (before any parallel `.map()` or `delayed()` call), and pass or scatter them to workers:
> 
> ```python
> # Serial regridder creation
> regridders = [xe.Regridder(ds, ds_out, 'bilinear') for ds in dataset_list]
> scattered = client.scatter(regridders)
> 
> # Parallel regridding
> futures = client.map(lambda ds, r: r(ds['var']), dataset_list, scattered)
> results = client.gather(futures)
> ```
> 
> Avoid initializing any `xESMF.Regridder` **inside** a function that is passed to `.map()` or any parallel task — this breaks ESMF internals.
>

tomvothecoder:
`xESMF` relies on `ESMF`/`ESMPy`, which is **not thread-safe** and uses **global state**, making it sensitive to concurrency issues. When using `Dask`, especially with `client.map`, it’s critical to ensure that:

* **Each xESMF task runs in its own dedicated process.**
* Tasks **must not share memory** or regridder instances.

If multiple xESMF operations run concurrently in the same process (e.g., through shared Dask workers), this can cause **segfaults**, **corrupted results**, or **silent hangs**.

### ✅ Recommended setup:

```python
Client(n_workers=N, threads_per_worker=1, processes=True)
```

Then use `client.map` or `client.submit` with isolated tasks — ideally one task per worker at a time.

### 💡 Tips:

* Precompute and reuse weights via `filename='weights.nc'` to avoid repeated construction overhead.
* Never share `Regridder` objects between tasks.
* Avoid the threaded scheduler (`processes=False`), which is unsafe for `xESMF`.

This pattern ensures stable and parallel-safe usage of `xESMF` in Dask workflows.

PR #967, opened 2025-04-15
Add U10 and CERES EBAF TOA v4.1 to polar maps

chengzhuzhang:
## Description

The CERES EBAF TOA variables are accidentally left out a while ago when transitioning to use CERES EBAF v4.1 from v4.0. This PR added back these variables, as well as surface wind speed U10 to compare with ERA5 data. 
Example output for these new variables are available [here](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.zhang40/tests/test_polar/viewer/polar/index.html).

- Closes #965 

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

If applicable:

- [ ] New and existing unit tests pass with my changes (locally and CI/CD build)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

PR #974, opened 2025-05-02
Creates a mask for 2d datasets before regridding

chengzhuzhang:
## Description
This PR is to address the regridding difference identified here:https://github.com/E3SM-Project/e3sm_diags/pull/964
lat_lon_HadISST_CL-SST-metrics-diff.ipynb:
- Shows differences in SST (Sea Surface Temperature) metrics (difference only shown in metrics table)
- v2 to v3 reference mean changes (18.698°C vs 18.777°C) due to regridding differences
- In v3, explicit masks are needed to prevent missing data from affecting results
- Provides possible solutions for handling masks with the xcdat regridder

This PR explicitly create a mask variable before regriding to prevent missing data affecting regridding results.

## Checklist

If applicable:

### DIFF
--- a/e3sm_diags/driver/utils/regrid.py
+++ b/e3sm_diags/driver/utils/regrid.py
@@ -391,18 +391,14 @@ def align_grids_to_lower_res(
         output_grid = ds_a_new.regridder.grid
         # Only create mask for 2D data (with no vertical dimension [zonal_mean_2d_*, meridional_mean_2d], or time dimension [annual_cycle_zonal_mean])
 
-        ds_b_new = _add_mask_for_2d(
-            ds_b_new, var_key, tool, ["lev", "plev", "z", "Z", "time", "T"]
-        )
+        ds_b_new = _add_mask(ds_b_new, var_key, tool)
         ds_b_regrid = ds_b_new.regridder.horizontal(
             var_key, output_grid, tool=tool, method=method
         )
         return ds_a_new, ds_b_regrid
 
     output_grid = ds_b_new.regridder.grid
-    ds_a_new = _add_mask_for_2d(
-        ds_a_new, var_key, tool, ["lev", "plev", "z", "Z", "time", "T"]
-    )
+    ds_a_new = _add_mask(ds_a_new, var_key, tool)
 
     ds_a_regrid = ds_a_new.regridder.horizontal(
         var_key, output_grid, tool=tool, method=method
@@ -411,10 +407,12 @@ def align_grids_to_lower_res(
     return ds_a_regrid, ds_b_new
 
 
-def _add_mask_for_2d(
-    ds: xr.Dataset, var_key: str, tool: str, dims_to_check: List[str]
-) -> xr.Dataset:
-    """Add a mask variable to the dataset if needed based on the tool and dimensions.
+def _add_mask(ds: xr.Dataset, var_key: str, tool: str) -> xr.Dataset:
+    """Add a mask variable to the dataset.
+
+    This function creates a mask variable for the specified variable key in
+    the dataset if the tool is "regrid2" (which supports 3D variables) or if
+    the variable is 2D with only spatial dimensions (e.g., "X" and "Y").
 
     Parameters
     ----------
@@ -424,25 +422,34 @@ def _add_mask_for_2d(
         The variable key.
     tool : str
         The regridding tool being used.
-    dims_to_check : List[str]
-        A list of dimensions to check for skipping mask creation.
 
     Returns
     -------
     xr.Dataset
         The modified dataset with the mask variable added if applicable.
     """
-    var_dims = ds[var_key].dims
+    ds_new = ds.copy()
+    var = ds_new[var_key]
+    var_dims = var.dims
 
-    if tool == "regrid2" or not any(dim in dims_to_check for dim in var_dims):
+    spatial_dims = {"x", "y", "lon", "lat", "longitude", "latitude"}
+    is_spatial_var = len(var_dims) == 2 and all(
+        str(dim).lower() in spatial_dims for dim in var_dims
+    )
+
+    if tool == "regrid2" or is_spatial_var:
         logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
-        ds["mask"] = xr.where(~np.isnan(ds[var_key]), 1, 0)
+
+        if "mask" in ds_new:
+            logger.warning("Overwriting existing 'mask' variable in the dataset.")
+
+        ds_new["mask"] = xr.where(~np.isnan(var), 1, 0)
     else:
         logger.debug(
             f"Skipping mask creation for variable {var_key} with dimensions {var_dims}"
         )
 
-    return ds
+    return ds_new
 
 
 def _are_grids_equal(ds_a: xr.Dataset, ds_b: xr.Dataset) -> bool:
--- a/e3sm_diags/driver/utils/regrid.py
+++ b/e3sm_diags/driver/utils/regrid.py
@@ -389,40 +389,60 @@ def align_grids_to_lower_res(
 
     if is_a_lower_res:
         output_grid = ds_a_new.regridder.grid
+        # Only create mask for 2D data (with no vertical dimension [zonal_mean_2d_*, meridional_mean_2d], or time dimension [annual_cycle_zonal_mean])
 
-        # Only create mask for 2D data (no vertical dimension [zonal_mean_2d_*, meridional_mean_2d], or time dimension [annual_cycle_zonal_mean])
-        var_dims = ds_b_new[var_key].dims
-        if tool == "regrid2" or not any(
-            dim in ["lev", "plev", "z", "Z", "time", "T"] for dim in var_dims
-        ):
-            logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
-            ds_b_new["mask"] = xr.where(~np.isnan(ds_b_new[var_key]), 1, 0)
-        else:
-            logger.debug(f"Skipping mask creation for 3D variable {var_key} with xesmf")
-
+        ds_b_new = _add_mask_for_2d(
+            ds_b_new, var_key, tool, ["lev", "plev", "z", "Z", "time", "T"]
+        )
         ds_b_regrid = ds_b_new.regridder.horizontal(
             var_key, output_grid, tool=tool, method=method
         )
-
         return ds_a_new, ds_b_regrid
 
     output_grid = ds_b_new.regridder.grid
-
-    # Only create mask for 2D data (no vertical dimension)
-    var_dims = ds_a_new[var_key].dims
-    if tool == "regrid2" or not any(
-        dim in ["lev", "plev", "z", "Z"] for dim in var_dims
-    ):
-        logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
-        ds_a_new["mask"] = xr.where(~np.isnan(ds_a_new[var_key]), 1, 0)
-    else:
-        logger.debug(f"Skipping mask creation for 3D variable {var_key} with xesmf")
+    ds_a_new = _add_mask_for_2d(
+        ds_a_new, var_key, tool, ["lev", "plev", "z", "Z", "time", "T"]
+    )
 
     ds_a_regrid = ds_a_new.regridder.horizontal(
         var_key, output_grid, tool=tool, method=method
     )
 
     return ds_a_regrid, ds_b_new
+
+
+def _add_mask_for_2d(
+    ds: xr.Dataset, var_key: str, tool: str, dims_to_check: List[str]
+) -> xr.Dataset:
+    """Add a mask variable to the dataset if needed based on the tool and dimensions.
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        The dataset to modify.
+    var_key : str
+        The variable key.
+    tool : str
+        The regridding tool being used.
+    dims_to_check : List[str]
+        A list of dimensions to check for skipping mask creation.
+
+    Returns
+    -------
+    xr.Dataset
+        The modified dataset with the mask variable added if applicable.
+    """
+    var_dims = ds[var_key].dims
+
+    if tool == "regrid2" or not any(dim in dims_to_check for dim in var_dims):
+        logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
+        ds["mask"] = xr.where(~np.isnan(ds[var_key]), 1, 0)
+    else:
+        logger.debug(
+            f"Skipping mask creation for variable {var_key} with dimensions {var_dims}"
+        )
+
+    return ds
 
 
 def _are_grids_equal(ds_a: xr.Dataset, ds_b: xr.Dataset) -> bool:
--- a/e3sm_diags/driver/utils/regrid.py
+++ b/e3sm_diags/driver/utils/regrid.py
@@ -408,36 +408,36 @@ def align_grids_to_lower_res(
 
 
 def _add_mask(ds: xr.Dataset, var_key: str, tool: str) -> xr.Dataset:
-    """Add a mask variable to the dataset.
+    """Add a mask variable to the dataset under specific conditions.
 
-    This function creates a mask variable for the specified variable key in
-    the dataset if the tool is "regrid2" (which supports 3D variables) or if
-    the variable is 2D with only spatial dimensions (e.g., "X" and "Y").
+    This function creates a mask variable for the specified variable in the
+    dataset if:
+        1. The regridding tool is "regrid2", which supports 3D mask variables, or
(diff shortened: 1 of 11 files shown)


chengzhuzhang:
Re-run [a full set run](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.zhang40/E3SMv3_regridding_mask_05022025/v3.LR.historical_0051/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1985-2014/viewer/index.html). The SST metrics are now closer to what we have in v2:
New:
Variables	Unit	Test_mean	Ref._mean	Mean_Bias	Test_STD	Ref._STD	RMSE	Correlation
SST global HadISST_CL	degC	20.256	18.674	1.583	8.178	9.558	1.054	0.992
SST global HadISST_PI	degC	20.256	18.964	1.292	8.178	8.952	1.232	0.991
SST global HadISST_PD	degC	20.256	18.787	1.469	8.178	9.561	1.082	0.992

V2:
Variables	Unit	Test_mean	Ref._mean	Mean_Bias	Test_STD	Ref._STD	RMSE	Correlation
SST global HadISST_CL	degC	20.256	18.698	1.559	8.178	9.536	1.054	0.992
SST global HadISST_PI	degC	20.256	18.978	1.279	8.178	8.933	1.232	0.991
SST global HadISST_PD	degC	20.256	18.807	1.45	8.178	9.543	1.082	0.992

V3:
Variables	Unit	Test_mean	Ref._mean	Mean_Bias	Test_STD	Ref._STD	RMSE	Correlation
SST global HadISST_CL	degC	20.256	18.777	1.48	8.178	9.464	1.055	0.992
SST global HadISST_PI	degC	20.256	19.058	1.199	8.178	8.853	1.233	0.991
SST global HadISST_PD	degC	20.256	18.885	1.372	8.178	9.47	1.082	0.992

chengzhuzhang:
@tomvothecoder I think this PR is in good shape, please review when get a chance, thanks!

pyproject.toml
```diff
@@ -218,7 +218,7 @@ testpaths = "tests/e3sm_diags"
 python_version = "3.10"
 check_untyped_defs = true
 ignore_missing_imports = true
-warn_unused_ignores = true
+warn_unused_ignores = false
```
tomvothecoder, line 221:
```suggestion
warn_unused_ignores = true
```

I think we should keep this because with new releases of `mypy`, old `# type: ignore` comments might no longer be relevant and we can remove them.

tests/e3sm_diags/driver/utils/test_regrid.py
```diff
@@ -283,7 +283,7 @@ def test_regrids_to_first_dataset_with_equal_latitude_points(self, tool):
         else:
             assert_identical(result_b, expected_b)
 
-    @pytest.mark.parametrize("tool", ("xesmf", "regrid2"))
+    @pytest.mark.parametrize("tool", ("regrid2",))  # Skip xesmf for this test
```
tomvothecoder, line 286: I'm curious why xesmf is removed from the tests? I noticed the test still passes with xesmf, which I would think we need to keep in the tests.

Also xCDAT ignores the `regrid_method` parameter when using `regrid2` since it is always `conservative`. We can raise a logger info message that the method will always be `conservative` and the `regrid_method` parameter is ignored. 

Or to make things cleaner, can even deprecate the option to use xCDAT's "regrid2" since we it is not used or tested in e3sm_diags v3.0.0.

e3sm_diags/driver/utils/regrid.py
```diff
@@ -383,13 +389,35 @@ def align_grids_to_lower_res(
...
+    # Only create mask for 2D data (no vertical dimension)
+    var_dims = ds_a_new[var_key].dims
+    if tool == "regrid2" or not any(
+        dim in ["lev", "plev", "z", "Z"] for dim in var_dims
+    ):
+        logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
+        ds_a_new["mask"] = xr.where(~np.isnan(ds_a_new[var_key]), 1, 0)
+    else:
+        logger.debug(f"Skipping mask creation for 3D variable {var_key} with xesmf")
+
```
tomvothecoder, line 420: Here's a refactored version of `align_grids_to_lower_res()` which extracts the common logic for creating the `mask` variable into a child function:

```python
def align_grids_to_lower_res(
    ds_a: xr.Dataset,
    ds_b: xr.Dataset,
    var_key: str,
    tool: REGRID_TOOLS,
    method: str,
    axis_to_compare: str = "Y",
) -> Tuple[xr.Dataset, xr.Dataset]:
    """Align the grids of two Dataset using the lower resolution of the two.

    Using the legacy logic, compare the number of latitude coordinates to
    determine if A or B has lower resolution:
      * If A is lower resolution (A <= B), regrid B -> A.
      * If B is lower resolution (A > B), regrid A -> B.

    This function creates a mask (where values aren't NaN) for the dataset being
    regridded to prevent NaN values from affecting nearby valid data points during
    regridding operations. The mask is added as a variable named "mask" to the dataset
    following the approach recommended by xESMF.

    Parameters
    ----------
    ds_a : xr.Dataset
        The first Dataset containing ``var_key``.
    ds_b : xr.Dataset
        The second Dataset containing ``var_key``.
    var_key : str
        The key of the variable in both datasets to regrid.
    tool : {"xesmf", "regrid2"}
        The regridding tool to use.
    method : str
        The regridding method to use. Refer to [1]_ for more information on
        these options.

        xesmf options:
          - "bilinear"
          - "conservative"
          - "conservative_normed"
          - "patch"
          - "nearest_s2d"
          - "nearest_d2s"

        regrid2 options:
          - "conservative"
    axis_to_compare : str
        The axis to use to compare resolutions to find the lower resolution
        of both variables, either "X" or "Y". by default "Y".

    Returns
    -------
    Tuple[xr.Dataset, xr.Dataset]
        A tuple of both DataArrays regridded to the lower resolution of the two.

    References
    ----------
    .. [1] https://xcdat.readthedocs.io/en/stable/generated/xarray.Dataset.regridder.horizontal.html
    .. [2] https://xesmf.readthedocs.io/en/latest/notebooks/Masking.html
    """
    ds_a_new = ds_a.copy()
    ds_b_new = ds_b.copy()

    # If grids are equal, then no regridding is required. This ensures no
    # performance cost is incurred.
    equal_grids = _are_grids_equal(ds_a_new, ds_b_new)
    if equal_grids:
        return ds_a_new, ds_b_new

    ds_a_new = _drop_unused_ilev_axis(ds_a_new)
    ds_b_new = _drop_unused_ilev_axis(ds_b_new)

    axis_a = xc.get_dim_coords(ds_a_new[var_key], axis=axis_to_compare)
    axis_b = xc.get_dim_coords(ds_b_new[var_key], axis=axis_to_compare)

    is_a_lower_res = len(axis_a) <= len(axis_b)

    if is_a_lower_res:
        output_grid = ds_a_new.regridder.grid
        ds_b_new = _add_mask_if_needed(
            ds_b_new, var_key, tool, ["lev", "plev", "z", "Z", "time", "T"]
        )
        ds_b_regrid = ds_b_new.regridder.horizontal(
            var_key, output_grid, tool=tool, method=method
        )
        return ds_a_new, ds_b_regrid

    output_grid = ds_b_new.regridder.grid
    ds_a_new = _add_mask_if_needed(ds_a_new, var_key, tool, ["lev", "plev", "z", "Z"])

    ds_a_regrid = ds_a_new.regridder.horizontal(
        var_key, output_grid, tool=tool, method=method
    )

    return ds_a_regrid, ds_b_new


def _add_mask_if_needed(
    ds: xr.Dataset, var_key: str, tool: str, dims_to_check: List[str]
) -> xr.Dataset:
    """Add a mask variable to the dataset if needed based on the tool and dimensions.

    Parameters
    ----------
    ds : xr.Dataset
        The dataset to modify.
    var_key : str
        The variable key.
    tool : str
        The regridding tool being used.
    dims_to_check : List[str]
        A list of dimensions to check for skipping mask creation.

    Returns
    -------
    xr.Dataset
        The modified dataset with the mask variable added if applicable.
    """
    var_dims = ds[var_key].dims

    if tool == "regrid2" or not any(dim in dims_to_check for dim in var_dims):
        logger.debug(f"Creating mask for {var_key} with dimensions {var_dims}")
        ds["mask"] = xr.where(~np.isnan(ds[var_key]), 1, 0)
    else:
        logger.debug(
            f"Skipping mask creation for variable {var_key} with dimensions {var_dims}"
        )

    return ds
```

tomvothecoder (commented):
Hi @chengzhuzhang, I left some questions and suggestions. This PR is coming together and almost done. Thanks.

pyproject.toml
chengzhuzhang: This flag is flipped for fixing pre-comming issue documented https://github.com/E3SM-Project/e3sm_diags/issues/977.

tomvothecoder: The unused type ignore warnings were resolved in #967 by deleting them. I committed my suggestion to flip it back to `true`. Let's see if it still fails.

tomvothecoder: I reverted addition of `# type: ignore` statements to align with `main` (https://github.com/E3SM-Project/e3sm_diags/pull/974/commits/2d7a20d9be4b7fdfb22010d9d4fb89033884ef3f) and now `pre-commit` passes.

e3sm_diags/driver/zonal_mean_2d_driver.py
```diff
@@ -203,6 +203,7 @@ def _get_avg_for_regridded_datasets(
 
     # Get the difference between the regridded variables and use it to
     # make sure the regridded variables have the same mask.
+
```
tomvothecoder, line 206:
```suggestion
```

e3sm_diags/driver/utils/regrid.py
```diff
@@ -383,20 +389,62 @@ def align_grids_to_lower_res(
...
+        A list of dimensions to check for skipping mask creation.
+
+    Returns
+    -------
+    xr.Dataset
+        The modified dataset with the mask variable added if applicable.
+    """
+    var_dims = ds[var_key].dims
+
+    if tool == "regrid2" or not any(dim in dims_to_check for dim in var_dims):
```
tomvothecoder, line 437:
```suggestion
    # Create a mask for xESMF if the variable is 2D (does not include specified dimensions)
    if tool == "xesmf" and not any(dim in dims_to_check for dim in var_dims):
```

Are we supposed to add the `mask` variable whenever the tool is `xesmf`, rather than `regrid2`?

chengzhuzhang: I think the logic is right without changing, assumming `regrid2` can take 3d var as input for regriding?

e3sm_diags/driver/utils/regrid.py
```diff
@@ -424,25 +422,34 @@ def _add_mask_for_2d(
...
+            logger.warning("Overwriting existing 'mask' variable in the dataset.")
+
+        ds_new["mask"] = xr.where(~np.isnan(var), 1, 0)
     else:
         logger.debug(
             f"Skipping mask creation for variable {var_key} with dimensions {var_dims}"
         )
 
-    return ds
+    return ds_new
```
tomvothecoder, line 452: I updated the logic of `_add_mask` to make it easier to understand. Now, it checks if the variable is a 2D spatial variable (if it is, it allows regridding using xESMF).

This is simpler than the old approach, which checked whether *any* of the variable’s dimensions weren't in a list of vertical and time dimensions. That method was harder to follow and skipped regridding with xESMF if any non-spatial dimension was present.

tests/e3sm_diags/driver/utils/test_regrid.py
```diff
@@ -627,3 +630,106 @@ def test_raises_error_for_dask_array(self):
...
+        )
+
+        with caplog.at_level(logging.WARNING):
+            result = _add_mask(ds, "var_2d", tool="xesmf")
+
+        # Ensure the "mask" variable is overwritten
+        np.testing.assert_array_equal(result["mask"].values, np.array([[1, 0], [1, 1]]))
+
+        # Ensure a warning is logged
+        assert "Overwriting existing 'mask' variable in the dataset" in caplog.text
```
tomvothecoder, line 735: I added unit tests to cover various cases with `_add_mask()`. 

It is easier to test `_add_mask()` directly because `align_grids_to_lower_res()` removes the `mask` variable by the end of regridding, which makes it harder to detect if a `mask` was used unless we write more comprehensive test to check for `np.nan`.

tomvothecoder (approved):
Hi Jill, I pushed some final suggestions. I think this PR is good to go if you agree with the changes.

Let me know what you think.

e3sm_diags/driver/utils/regrid.py
tomvothecoder: Also, it raises a `logger.warning` is an existing `mask` variable will be overwritten

e3sm_diags/driver/utils/regrid.py
```diff
@@ -383,20 +389,69 @@ def align_grids_to_lower_res(
...
+    -------
+    xr.Dataset
+        The modified dataset with the mask variable added if applicable.
+    """
+    ds_new = ds.copy()
+    var = ds_new[var_key]
+    var_dims = var.dims
+
+    spatial_dims = {"x", "y", "lon", "lat", "longitude", "latitude"}
+    is_spatial_var = len(var_dims) == 2 and all(
```
chengzhuzhang, line 436: I feel this is a little risky than the old logic, for cases when excessive dims present? I'm not sure if those were completely dropped at some point. I can test a full run again to see if all variables are produced.

tomvothecoder: I originally understood the logic for xESMF regridding as: ignore any dimensions that are Z (vertical) or T (time), which effectively means we expect to regrid a 2D variable with just X and Y (horizontal spatial dimensions).

As far as I can recall, I haven't encountered any variables with more than these four dimensions (X, Y, Z, T). If such variables do exist and we want to proceed with regridding them anyway, this logic would end up ignoring those variables entirely.

Also, based on what I'm reading, xESMF does support regridding variables with more than two dimensions by broadcasting. It applies the 2D regridding independently across each slice along the non-spatial dimensions. Do we want to continue regridding variables that have >2 dimensions minus Z and T?

tomvothecoder: I originally understood the logic for xESMF regridding as: ignore any dimensions that are Z (vertical) or T (time), which effectively means we expect to regrid a 2D variable with just X and Y (horizontal spatial dimensions).

As far as I can recall, I haven't encountered any variables with more than these four dimensions (X, Y, Z, T) but there might be some. If such variables do exist and we want to proceed with regridding them anyway, this logic would end up ignoring those variables entirely which is problematic.

Also, based on what I'm reading, xESMF does support regridding variables with more than two dimensions by broadcasting. It applies the 2D regridding independently across each slice along the non-spatial dimensions. Do we want to continue regridding variables that have >2 dimensions minus Z and T? If so, then we'll need to revert the logic back to the old one.

chengzhuzhang: I think there are cosp related dimensions, but not sure if those are dropped in the course of data ingection 

xESMF does support 3Dvar as our original implmentation, but doesn't support 3d masks, as I understand. Here is the code line that throw errors when passing in 3d mask: https://github.com/pangeo-data/xESMF/blob/30e3ecb094d39a0c6f1edb22b60d8d608a19f7d9/xesmf/backend.py#L138.  Before xESMF relaxing this constrain, we can only have more accurate regridding for 2d vars, at this point.

chengzhuzhang: I do like the refactored and refined code from your change. Could you only revert the logic that handles exclusion of dims? Thanks!

tomvothecoder: Yeah sounds good!

e3sm_diags/driver/utils/regrid.py
```diff
@@ -408,36 +408,36 @@ def align_grids_to_lower_res(
...
 
-    spatial_dims = {"x", "y", "lon", "lat", "longitude", "latitude"}
-    is_spatial_var = len(var_dims) == 2 and all(
-        str(dim).lower() in spatial_dims for dim in var_dims
-    )
+    dims_to_skip = {"lev", "plev", "z", "time", "t"}
+    has_skipped_dims = any(str(dim).lower() in dims_to_skip for dim in var_dims)
 
-    if tool == "regrid2" or is_spatial_var:
+    if tool == "regrid2" or not has_skipped_dims:
```
tomvothecoder, line 440: Revert the logic for checking dimensions back to the old logic in [7bd8f29 (#974)](https://github.com/E3SM-Project/e3sm_diags/pull/974/commits/7bd8f290993f4a3e8edad2341cd23b57cbaf5ced)

chengzhuzhang: Thank you!

PR #984, opened 2025-06-02
Fix layout for sub-plot saved in pdf format

chengzhuzhang:
## Description

When saving sub panel as pdf format, problems are reported about cropped axis, lables. This PR aims to fixes sets that have subplot saving problems.
- Closes #970 meridional and zonal mean 2d plots, closes #294 (for enso) and other sets.
- tropcial_subseasonal: fixing padding border 
- aerosol_aeronet: fixing padding border and plotting
- enso_diags: fixing parser to make command-line work and fixing padding border
- QBO: fixing fig position and padding border
## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

If applicable:

- [ ] New and existing unit tests pass with my changes (locally and CI/CD build)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

Issue #970, opened 2025-04-25
[Bug]: Pressure-latitude zonal mean subplots have tight left margin and y-axis name partially clipped


wlin7:
### What happened?

The  pdf subplots of pressure-latitude zonal means plots have y-axis name `pressure `partially clipped. Allowing some left margin will help. This probably happens to all zonal-pressure subplots. The full plot (with all three panels are good). Here are two such subplots: [T](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.wlin/E3SMv3/v3.LR.historical_0101/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1985-2014/zonal_mean_2d/ERA5/ERA5-T-ANN-global.2.pdf) and [U](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.wlin/E3SMv3/v3.LR.historical_0101/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1985-2014/zonal_mean_2d/ERA5/ERA5-U-ANN-global.1.pdf) vs ERA5)

![Image](https://github.com/user-attachments/assets/ded0469f-309b-43ba-9a43-e9fa8cd2cb0a)

![Image](https://github.com/user-attachments/assets/6262f78a-6d11-4add-b252-c7b82d00ebbc)

### What did you expect to happen? Are there are possible answers you came across?

Allow some left margin to the left of y-axis name

### Minimal Complete Verifiable Example (MVCE)

```python

```

### Relevant log output

```python
N/A
```

### Anything else we need to know?

N/A

### Environment

produced by standard zppy workflow with recent or latest e3sm_unfied

chengzhuzhang:
Update: For immediate correction, to update command line option: --figsize '8.5' '11.0' should change the layout of the pdf file and potentially get the y label within the figure. After some experimenting, @wlin7 found out that to use "10","12.94" as figsize works out okay. For the long term, we need to fix this problem in the code:
the problem is in the zonal_mean_2d plots where the left margin is too tight, causing the y-axis label "pressure (mb)" to be partially clipped.  The issue is in the DEFAULT_PANEL_CFG values in e3sm_diags/e3sm_diags/plot/utils.py, where the left margin (0.1691) isn't providing enough space for the y-axis label.

It worked fine for png because rasterization may alters text positioning..

The command line to use:
```
e3sm_diags zonal_mean_2d --no_viewer --reference_data_path 
[...]

PR #987, opened 2025-06-11
Revert "Fix layout for sub-plot saved in pdf format"

chengzhuzhang:
Reverts E3SM-Project/e3sm_diags#984

When I tested again after merging the PR, the pdf plots all return as eampty pdfs. Before merging the PR I might have tested an early commit before the refactor commit.

### DIFF
--- a/e3sm_diags/driver/arm_diags_driver.py
+++ b/e3sm_diags/driver/arm_diags_driver.py
@@ -278,11 +278,6 @@ def _run_diag_annual_cycle(parameter: ARMDiagsParameter) -> ARMDiagsParameter:
                     vars_funcs = _get_vars_funcs_for_derived_var(ds_ref, var)
                     target_var = list(vars_funcs.keys())[0][0]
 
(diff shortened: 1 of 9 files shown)


tomvothecoder:
Good idea to try to keep this commit because it reduces the repeated logic to make the code cleaner. The empty pdf might be a simple fix.

chengzhuzhang:
> I will work on this today to see why plots are empty. Can you provide the minimum Python script? Otherwise I'll adapt the command line [here](https://github.com/E3SM-Project/e3sm_diags/issues/970#issuecomment-2831089971).

I don't have the Python script, but only used those command lines to produce results for each set being changed. I think you can just use the aerosol_aeronet line because it runs faster than others. The refactored version may impact all sets. That's why I was hesitating to make more refactor when working on this PR...but if we can quickly find a simple fix, that would be great.

chengzhuzhang:
main is fixed by #989, no longer need to revert

PR #1042, opened 2026-03-10
Fix Python 3.14 stall due to repeated climo opens without close

### LINKED ISSUE #1040: [Bug]: Multiprocessing fork context causes thread deadlock
### What happened?

The original implementation used fork (Linux default before Python 3.14) as the multiprocessing start method (via Dask). Recently when testing with two long runs on Chrysalis, two jobs were failed and potentially caused with the fork context: ex. /lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1181439.fork. 

Explained by copilot:

:fork copies only the calling thread into the child process, but inherits
  all mutex state from the parent. When the parent process has background threads holding locks — common with scientific libraries like OpenMP, MKL, or Python's own logging/GIL
  infrastructure — the forked child inherits those locked mutexes with no thread to ever unlock them, causing a deadlock. This was observed in two jobs with the fork context.

### What did you expect to happen? Are there are possible answers you came across?

I'm experimenting to switch to use `spawn` or `forkserver` as alternative multiprocessing context options, with later as the default for Linux Python 3.14

### Minimal Complete Verifiable Example (MVCE)

```python
/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.bash
```

### Relevant log output

```python

```

### Anything else we need to know?

_No response_

### Environment

/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1181439.fork

tomvothecoder:
## Summary

Fix the Python 3.14 climo open stall by removing the raw `netCDF4` workaround and making dataset ownership explicit in the xCDAT/xarray path.

This change keeps `xc.open_mfdataset()` as the supported climo open path, then loads and closes file-backed datasets deterministically so repeated opens no longer depend on GC timing.

Closes #1040 

## What Changed

- Removed the `RAW_NETCDF4_CLIMO_OPEN` runtime path from `dataset_xr.py`
- Kept climatology opens on `xc.open_mfdataset()` only
- Refactored climo and time-series dataset loading to:
  - keep the original lazily opened dataset separate from the loaded result
  - close file-backed datasets in `finally` blocks on both success and failure
- Updated `_subset_vars_and_load()` to load subsets into memory and explicitly close backing datasets
- Switched `_get_global_attr_from_climo_dataset()` to use `_open_climo_dataset()` and removed the leftover direct `netCDF4` attr-read helper
- Added/updated regression tests for:
  - scalar-time retry behavior
  - repeated climo opens
  - close-on-success / close-on-error behavior
  - global attr reads through `_open_climo_dataset()`
- Removed temporary debug hang instrumentation and debug-only env overrides from the repro helpers

## Why

On Python 3.14, repeated climo opens could stall even in serial mode. The issue appeared to be tied to delayed cleanup of file-backed xarray/xCDAT datasets, not to multiprocessing itself. This PR fixes that by making cleanup deterministic while preserving the existing xCDAT-based workflow.

## Checklist

If applicable:


tomvothecoder (commented):
@chengzhuzhang I was able to reproduce the serial stalling on Chrysalis (both compute and login nodes), consistently around the ALBEDO variable using this [min script](https://raw.githubusercontent.com/E3SM-Project/e3sm_diags/1d6dae6ea61b5b25255d57c2f0f6b93afd8a139c/auxiliary_tools/debug/1040-py314-hang/qa.py). After isolating the code path, the stall appears to happen while opening a climo dataset with xarray/xcdat just to read the global `yrs_averaged` attribute.

I pushed a potential fix that reads `yrs_averaged` directly with netCDF4 instead of calling `_open_climo_dataset()` for that lookup. Could you check out this branch, install it in your environment, and verify whether the stalling still occurs?

More info in the above comment: https://github.com/E3SM-Project/e3sm_diags/pull/1042#issuecomment-4040622125


## Notes

- The way climo files are accessed in zppy bash scripts (for example via symlinks) may contribute to the stall.
- There may be NetCDF/xarray resource contention exposed by Python 3.14.
- We also still rely on garbage collection for some dataset cleanup (not always explicit ds.close()).

## Follow ups
- [x] Test on another machine to rule out machine/filesystem-specific behavior. -- **Jill confirmed that e3sm_diags works fine on Perlmutter [here](https://github.com/E3SM-Project/e3sm_diags/issues/1040#issuecomment-4042254178)**

auxiliary_tools/debug/1040-py314-hang/parallel/qa.py
```diff
@@ -0,0 +1,254 @@
...
+    for item in os.environ.get("E3SM_DIAGS_REPRO_SEASONS", "ANN,DJF").split(",")
+    if item.strip()
+]
+REPRO_REGIONS = [
+    item.strip()
+    for item in os.environ.get(
+        "E3SM_DIAGS_REPRO_REGIONS", "polar_S,polar_N"
+    ).split(",")
+    if item.strip()
+]
```
tomvothecoder:
```suggestion
NUM_WORKERS = 8
os.environ["RAW_NETCDF4_CLIMO_OPEN"] = "1"
```

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -342,17 +374,53 @@ def _get_global_attr_from_climo_dataset(
...
-            attr_val = ds.attrs.get(attr)
+            # Read global metadata without opening full xarray/xcdat datasets.
+            _log_hang_debug(f"reading global attr via netCDF4: {filepath}")
+
+            try:
+                attr_val = self._read_global_attr_with_netcdf4(filepath, attr)
+            except Exception:
+                _log_hang_debug(
+                    f"netCDF4 global attr read failed; opening climo dataset: {filepath}"
+                )
```
tomvothecoder:
```suggestion

```

e3sm_diags/parameter/core_parameter.py
```diff
@@ -356,8 +369,21 @@ def _set_name_yrs_attrs(
         time_selection : TimeSelection | None
             The optional time slice or season.
         """
```
tomvothecoder:
```suggestion
        """
        t0 = time.monotonic()
        _log_hang_debug(f"start _set_name_yrs_attrs selection={time_selection}")
```

tomvothecoder (commented):
Hey @chengzhuzhang, I was able to reproduce the parallel stalling using this [qa.py](https://github.com/E3SM-Project/e3sm_diags/pull/1042/changes#diff-6cbaa1a65236f5f113c04465d127fa7a809236cda401ad4209f8780872481ad9) script, including with `fork`/`forkserver` and `num_workers=1`/`num_workers=8`.

Could you install my branch and try a parallel run on your side?

## Current Finding

The issue seems to be in the climatology file open path in `Dataset._get_climo_dataset()`. The normal xarray/xCDAT filepath-based open path can stall, while a raw `netCDF4`-backed open path completed successfully in my shortened parallel min script.

I suspect repeated climatology file opens, **especially repeated reopening of the same .nc files** through the xarray/xCDAT filepath-based backend, are not releasing resources cleanly and eventually block the open path.

## Patch To Test

I updated `_get_climo_dataset()` so climo datasets can be opened through `netCDF4` first and then wrapped as an xarray Dataset. 

This worked as a workaround in my min script using either `fork` or `forkserver`.

## Next Steps

- [ ] Confirm whether this patch resolves the parallel stalling in your full workflow
- [ ] Add explicit dataset lifecycle management (ensure `ds.close()` is always called, including underlying `nc.close()` if using `NetCDF4Dataset`)
- [ ] If stalling is resolved, evaluate whether to keep this workaround or revert to the standard xarray/xCDAT open path with proper cleanup
- [ ] Decide if we want to keep using `forkserver` or revert to `fork` -- leaning `forkserver` as it seems more future-proof and robust, considering Python 3.14+ also switches to it by default

### Explore

- [ ] Validate whether explicit `ds.close()` alone makes the original open path stable under Dask bag parallelism
- [ ] Test `engine="h5netcdf"` as an alternative backend (lower overhead and reduced HDF5/libnetcdf locking under parallel reads)
- [ ] Compare behavior across:
  - [ ] `netcdf4` + explicit close
  - [ ] `h5netcdf` + explicit close
- [ ] Identify whether the root cause is:
  - file descriptor accumulation
  - HDF5 contention
  - repeated open frequency within tasks

chengzhuzhang:
Looking at this xarray issue https://github.com/pydata/xarray/issues/11241 , it doesn't look like xarray fully support py3.14. I think we can probably comfortably stay with py3.13.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -348,8 +347,13 @@ def _get_global_attr_from_climo_dataset(
...
-            attr_val = ds.attrs.get(attr)
+            ds_climo = self._open_climo_dataset(filepath)
+
+            try:
+                attr_val_raw = ds_climo.attrs.get(attr)
+                attr_val = str(attr_val_raw) if attr_val_raw is not None else None
+            finally:
+                ds_climo.close()
 
         return attr_val
```
tomvothecoder: Updates to `_get_global_attr_from_climo_dataset()` to explicitly close dataset.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -588,22 +592,25 @@ def _get_climo_dataset(self, season: str) -> xr.Dataset:
...
 
-        ds = squeeze_time_dim(ds)
-        ds = self._subset_vars_and_load(ds, self.var)
+            ds_work = squeeze_time_dim(ds_work)
+            ds_loaded = self._subset_vars_and_load(ds_work, self.var)
 
-        return ds
+            return ds_loaded
+        finally:
+            ds_open.close()
```
tomvothecoder: Updates to `get_time_sliced_dataset()` to explicitly close dataset.

tomvothecoder (commented):
Hey @chengzhuzhang, I've pushed changes to replace the raw `netCDF4` climo workaround with explicit xarray/xCDAT dataset cleanup.


**Can you test on your end with Python 3.13 and 3.14? If they are all successful, we should compare performance and then we should also check if all of the plots produced as usual.**

### Changes

TL;DR: this removes the raw netCDF4 fallback and fixes the Python 3.14 climo hang by making xCDAT/xarray dataset cleanup deterministic in dataset_xr.py. Repeated climo opens now stay on xc.open_mfdataset(), load needed data into memory, and explicitly close file-backed datasets on both success and error paths.

**NOTE: I've reverted `forkserver` back to `fork` to remove additional variables that might influence performance. More info [here](https://github.com/E3SM-Project/e3sm_diags/pull/1042#discussion_r3002459008), [commit](https://github.com/E3SM-Project/e3sm_diags/pull/1042/changes/bc33613a005e49c8094dc3968685eb029462bb79)**

### Results

| Run | Python  | Time | Status | Notes |
|---|---|---|---|---|
| Serial min script | 3.13 | 1933 secs | No stall ✅ | |
| Serial min script | 3.14 | 1952 secs | No stall ✅ | |
| Parallel min script | 3.13 | 139.5 secs | No stall ✅ | |
| Parallel parallel script | 3.14 | 136.7 secs | No stall ✅ | |
| Serial `sbatch` run | 3.13 | 4hr | No stall ✅  | Timed out (expected with serial) |
| Serial `sbatch` run | 3.14 | 4hr | No stall ✅  | Timed out (expected with serial) |
| Parallel `sbatch` run | 3.13 | 49m10s | No stall ✅  | |
| Parallel `sbatch` run | 3.14 | 48m22s | No stall ✅  | |

- Min scripts only test `sets=['polar']`, `variables=T`, `seasons=['ANN']`, `regions=['polar_S', 'polar_N']`, and `plevs=[200.0]`
- _`num_workers=8` for parallel runs._

### Logs
Root Dir: `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts`

| Run | Python | Log |
|---|---|---|
| Serial `sbatch` run | 3.13 | `e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1206454` |
| Serial `sbatch` run | 3.14 | `e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o120455` |
| Parallel `sbatch` run | 3.13 | `e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1206453` |
| Parallel `sbatch` run | 3.14 | `e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1206452` |


### Output Comparison
In Progress

Python 3.13
- Log: `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1207146`
- Results Dir: `/lcrc/group/e3sm/public_html/ac.tvo/1040-py314-hang-tom-py313/`

Python 3.14
- Log: `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1207147`
- Results Dir: `/lcrc/group/e3sm/public_html/ac.tvo/1040-py314-hang-tom-py314/`

tomvothecoder:
> The last check is to make sure both runs produced the same results with no missing plots. Once we confirm everything is good to go, we can merge this PR to unblock Python 3.14 for E3SM Unified. We can do further testing during the RC period too.

This branch looks ready to merge.

The Python 3.14 run completed successfully with 8 workers, matched `main` exactly across all 1882 PNGs, and showed no output regression from the explicit dataset-close change.

The only PNG differences appeared in the Python 3.13 branch run. Those differences are extremely small and look like pixel-level noise, not meaningful plot changes. They correlate more strongly with environment and package differences, especially `numpy=2.26` compared to `numpy=2.4.3` in the other branches, rather than with the code change itself.

## Run comparison

| Python | Branch | Log | Results dir | Web | Completion time |
|---|---|---|---|---|---|
| 3.13 | `main` | `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1207152` | `/lcrc/group/e3sm/public_html/ac.tvo/1040-py314-hang-tom-py313-main/` | [link](https://web.lcrc.anl.gov/public/e3sm/ac.tvo/1040-py314-hang-tom-py313-main/) | 49m23s |
| 3.13 | `bug/1040-py314-hang-tom` | `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1207148` | `/lcrc/group/e3sm/public_html/ac.tvo/1040-py314-hang-tom-py313/` | [link](https://web.lcrc.anl.gov/public/e3sm/ac.tvo/1040-py314-hang-tom-py313/) | 48m52s |
| 3.14 | `bug/1040-py314-hang-tom` | `/lcrc/group/e3sm/ac.zhang40/zppy_example_v3.2.0/v3.LR.historical_0051/post/scripts/e3sm_diags_atm_monthly_180x360_aave_model_vs_obs_1985-2014.o1207149` | `/lcrc/group/e3sm/public_html/ac.tvo/1040-py314-hang-tom-py314/` | [link](https://web.lcrc.anl.gov/public/e3sm/ac.tvo/1040-py314-hang-tom-py314/) | 49m40s |

**Takeaway:** all three 8-worker parallel runs completed in essentially the same time window.

## Output comparison

| Comparison | PNG count match | Missing/extra vs `main` | Identical PNGs | Different PNGs | Max diff fraction | Interpretation |
|---|---:|---:|---:|---:|---:|---|
| `py313_main` vs `py314` | Yes (`1882/1882`) | `0 / 0` | `1882` | `0` | `0.000000000000` | Exact match |
| `py313_main` vs `py313` | Yes (`1882/1882`) | `0 / 0` | `1845` | `37` | `0.000097920380` | Tiny pixel-level noise only |
| `py313` vs `py314` | Yes (`1882/1882`) | n/a | `1845` | `37` | `0.000097920380` | Same tiny noise pattern |


- [Branch Comparison Summary CSV
](https://github.com/E3SM-Project/e3sm_diags/blob/bug/1040-py314-hang-tom/auxiliary_tools/debug/1040-py314-hang/parallel/compare_outputs_summary.csv)
- [Diffs-Only Output Comparison CSV](https://github.com/E3SM-Project/e3sm_diags/blob/bug/1040-py314-hang-tom/auxiliary_tools/debug/1040-py314-hang/parallel/compare_output_diff_artifacts.csv)

## Environment signal

| Package / signal | `py313_main` | `py313` | `py314` | Notes |
|---|---:|---:|---:|---|
| Python | `3.13.12` | `3.13.12` | `3.14.3` | `py313_main` and `py314` still matched exactly |
| NumPy | `2.4.3` | `2.2.6` | `2.4.3` | strongest correlation with PNG differences |
| Xarray | `2026.2.0` | `2025.10.1` | `2025.10.1` | differs, but does not align with output mismatch as clearly as NumPy |
| Other package diffs | several | several | several | `13` package-level diffs total across envs |

- [Full environment differences CSV](https://github.com/E3SM-Project/e3sm_diags/blob/bug/1040-py314-hang-tom/auxiliary_tools/debug/1040-py314-hang/parallel/environment_dependency_differences.csv)

## Key points

- All three runs produced the same `1882` PNGs, with no missing files and no size mismatches.
- `py314` matches `py313_main` exactly, so the explicit dataset-close change does not show an output regression in this test.
- The only differences are the same `37` PNGs in `py313` relative to both `py313_main` and `py314`.
- Those diffs are tiny, about `0.0098%` max changed pixels, and the Pillow artifacts look like visual noise rather than meaningful plot changes.
  - [Example diff](https://web.lcrc.anl.gov/public/e3sm/ac.tvo/1040-py314-hang-tom-diffs/py313_vs_py314/model_vs_obs_1985-2014_units/polar/CERES-EBAF-TOA-v4.1/ceres_ebaf_toa_v4.1-FLUTC-DJF-polar_N/diff_raw.png)
  - [Diffs directory](https://web.lcrc.anl.gov/public/e3sm/ac.tvo/1040-py314-hang-tom-diffs/)
- This points away from Python 3.14 as the cause, since `py313_main` and `py314` use different Python versions but still produce identical outputs.
- The strongest environment correlation is `numpy`: `py313_main` and `py314` use `2.4.3`, while `py313` uses `2.2.6`.
  - `auxiliary_tools/debug/1040-py314-hang/parallel/environment_dependency_differences.csv`

## Bottom line

There is no evidence here of a real regression from the explicit dataset-close change.

The Python 3.14 branch run is clean, and the small `py313` PNG differences are most likely environment-related noise rather than a code-change regression.

PR #903, opened 2024-12-06
 [DevOps]: Add complete-run tests with metric and image checks

tomvothecoder:
## Description

Adds HPC complete-run regression validation for NetCDF outputs and PNG plots.

- Closes #894
- Closes #563
- Replaces the comparison notebook with a reusable Python comparison workflow.
- Adds baseline promotion, environment provenance, and numerical/image checks.
- Documents Layer 4 complete-run validation.
- Validated against `main`: 1,571 NetCDF files and 475 images matched with zero mismatches.

## Todos

- [x] Adapt regression testing scripts.
- [x] Generate and promote a validated `main` baseline.

## Checklist

- [x] I have performed a self-review of my own code
- [x] New and existing tests pass with my changes
- [x] I have added tests that prove my feature works
- [x] I have made corresponding changes to the documentation

PR #824, opened 2024-07-19
make --closedcontourcmd resolution specific

chengzhuzhang:
## Description

TC detection threshold should be model data resolution depended, currently the pre-processing script is using a threshold for ne120, resulted small number of detected TC events in ne30 simulations. Now update to add resolution specific thresholds.
Note: this fix should bring into cdat migration branch, and zppy implementation.

The original closed contour criteria (--closedcontourcmd "PSL,300.0,4.0,0;_AVG(T200,T500),-0.6,4,0.30"):
requiring an increase in PSL of at least 300 Pa (3 hPa) within 4 Degree of the candidate node,
and a decrease in 200  and 500 hPa average air temperature of 0.6 K within 4 Degree of the node within 0.3 degree of the candidate with maximum air temperature.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

If applicable:

- [ ] New and existing unit tests pass with my changes (locally and CI/CD build)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have noted that this is a breaking change for a major release (fix or feature that would cause existing functionality to not work as expected)

PR #1048, opened 2026-04-01
[Bug]: Prevent Xarray NetCDF3 climo read lock stalls

### LINKED ISSUE #1066: lat_lon_land run with multiprocessing=True hangs indefinitely at end of run (never exits, no viewer)
### What happened?

Reported by @xuezhengllnl  A `lat_lon_land` model-vs-model run with `multiprocessing=True` / `num_workers=32` **hangs indefinitely at the end of the run**. All variables and plots are generated successfully (~362 land variables × 5 seasons, ~1,582 plots, ~72 min), but the driver then produces **no further output and never exits** — no viewer/`index.html` is created. The process sits idle until SLURM kills it at the wall-clock limit.

This is **reproducible**: it happened across 4 independent runs, each going completely silent immediately after the **final** variable `LITTERC_HR` (last log line is always `Plot saved in: .../C Flux/...-LITTERC_HR-SON-global.png`), then hanging for hours until `CANCELLED ... DUE TO TIME LIMIT`.

**It is not resource exhaustion.** The node is `--exclusive` with 253 GB RAM. Node memory was sampled every 30 s and stayed **flat at ~25 GB used / ~221 GB free throughout, including during the hang** — so no OOM and no swapping. It is an idle deadlock, not slow compute.

### Likely-related error

Several variables raise a `ValueError` from unit parsing that is caught per-diagnostic (`core_parameter._run_diag`) and logged as ERROR, after which processing continues for other variables — but the run still deadlocks at the end:

```
[ERROR]: core_parameter.py(_run_diag:418) >> Error in e3sm_diags.driver.lat_lon_land_driver
  File ".../e3sm_diags/parameter/core_parameter.py", line 415, in _run_diag
    single_result = module.run_diag(self)
  File ".../e3sm_diags/driver/lat_lon_land_driver.py", line 13, in run_diag
    return base_run_diag(parameter)
  File ".../e3sm_diags/driver/lat_lon_driver.py", line 80, in run_diag
    ds_test = test_ds.get_climo_dataset(var_key, time_selection)
  ...
  File ".../e3sm_diags/derivations/derivations.py", line 1664, in <lambda>
    [(("FCH4TOCO2",), lambda v: convert_units(v, target_units="g*/m^2/day"))]
  File ".../e3sm_diags/derivations/utils.py", line 123, in convert_units
    original_udunit = cf_units.Unit(var_new.attrs["units"])
  File ".../cf_units/__init__.py", line 736, in __init__
    raise value_error from None
ValueError: [UT_UNKNOWN] Failed to parse unit "gC/m2/s"
```

Affected variables (source `units = "gC/m2/s"`, which udunits/cf_units cannot parse — it expects e.g. `g/m2/s`): **`CH4PROD`, `FCH4`, `FCH4TOCO2`** (and similar C-flux fields). Note `LITTERC_HR`, the last variable processed before every hang, is in the same `C Flux` group.

### Two issues, possibly linked

1. **Deadlock on exit with `multiprocessing=True`.** When a per-variable task raises (and/or on the final task), the multiprocessing pool does not appear to join/collect cleanly, so the main process blocks forever instead of finishing and writing the viewer. This makes the whole run unusable even though every plot was produced.
2. **`convert_units` cannot handle `"gC/m2/s"`.** Should `convert_units` normalize/strip the species prefix (`gC` → `g`) or fail more gracefully, so these ELM carbon/methane flux variables convert instead of erroring?

### Environment

- e3sm_diags **v3.2.0** (via e3sm-unified 1.13.0)
- Python 3.13, Linux, SLURM (Chrysalis)
- `sets_to_run = ['lat_lon_land']`, `run_type = model_vs_model`
- `multiprocessing = True`, `num_workers = 32`, exclusive node (253 GB RAM)

### Workaround being tested

Reducing `num_workers` / setting `multiprocessing = False`, and correcting the `gC/m2/s` unit strings on the input data.

tomvothecoder:
## Description

Prevents intermittent multiprocessing stalls when Xarray >=2026.01.0 opens read-only NetCDF3 climatology files through the netCDF4 backend.

The failure is timing- and workload-dependent. It reproduces in the ATM monthly 180x360 workflow with Xarray 2026.7.0 on Python 3.14.1–3.14.4 and in the `lat_lon_land` workflow on Python 3.13.14 ([details](https://github.com/E3SM-Project/e3sm_diags/pull/1048#issuecomment-5160950489)). This confirms that the issue is not limited to Python 3.14 or the `fork` multiprocessing context.

The fix:

- resolves climatology inputs and inspects their NetCDF data models
- uses `lock=False` only when all inputs are read-only NetCDF3 files
- preserves Xarray's default locking for NetCDF4/HDF5, mixed, unreadable, or unknown inputs
- reads global attributes directly from NetCDF metadata
- loads selected subsets before closing file-backed datasets
- reuses one seasonal climatology open for land and sea masks
- adds Python 3.14 support to `pyproject.toml`, conda environments, and the GitHub Actions build matrix

The full investigation, test matrices, performance results, and root-cause interpretation are documented in the [final report](https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/final_report.md).

Closes #1066.  
Related to #1040.

Upstream context:

- https://github.com/pydata/xarray/pull/10788
- https://github.com/pydata/xarray/issues/824
- https://github.com/pydata/xarray/issues/10843
- https://github.com/E3SM-Project/zppy/discussions/802#discussioncomment-16415585

## Validation

- With the workaround enabled:
  - all 10 Python environments completed the minimum reproduction and full ATM workflow
  - 50/50 minimum-test iterations completed
  - 10/10 full runs generated viewers
- Without the workaround, Python 3.14.1–3.14.4 stalled in both test modes, and full jobs reached the four-hour timeout.
- Python 3.13.14 and 3.14.6 produced identical NetCDF data, canonical JSON, and PNG pixels with and without the workaround.
- The Python 3.13.14 `lat_lon_land` test with Xarray 2026.7.0 and `fork` completed all 1,697 plots, matching Xarray 2025.12.0.
- No performance regression was established:
  - ATM cohorts differed by 1 minute 28 seconds (3.2%)
  - the fixed land-set run completed in 1:14:02 versus a 1:16:02 baseline
- `make test`
- CI/CD unit tests

## Checklist

If applicable:

- [x] New and existing unit tests pass locally and in CI/CD
- [x] I have added tests that prove the fix is effective
- [x] I have added an investigation report for the user-facing behavior change
- [ ] I have noted that this is a breaking change for a major release


tomvothecoder:
> @tomvothecoder thank you for starting exploring again. I checked my python 3.14 env and found that the xarray version i use is 2025.10.1, while in Ryan's zppy env, xarray version is xarray 2026.2.0. Could you check which xarray version is used in your test? I'm wondering if difference in underling package version play a role here.

Thanks for checking. My Python 3.14 environment [here](https://github.com/E3SM-Project/e3sm_diags/pull/1042#issuecomment-4151213828) also used the older Xarray and UXarray versions, but it succeeded without stalling.

I have two new runs going for Python 3.14 with the latest dependencies ([comment above](https://github.com/E3SM-Project/e3sm_diags/pull/1048#issuecomment-4173567554)): one on `main` and one on this branch. Will update with the results soon.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -350,13 +348,22 @@ def _get_global_attr_from_climo_dataset(
...
+
+                with netCDF4.Dataset(filepath, mode="r") as ds_climo:
+                    attr_val_raw = (
+                        ds_climo.getncattr(attr) if attr in ds_climo.ncattrs() else None
+                    )
                 attr_val = str(attr_val_raw) if attr_val_raw is not None else None
-            finally:
-                ds_climo.close()
+            except OSError:
+                pass
```
tomvothecoder, line 459: Optimized this code to grab just one attribute from a climatology file using the netCDF4 package (already a dependency), instead of using xarray open_mfdataset.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -676,24 +677,64 @@ def _open_climo_dataset(self, filepath: str) -> xr.Dataset:
...
 
+        logger.info(
+            "Climo backend %s start pid=%s var=%s filepath=%s",
+            "open_mfdataset",
+            os.getpid(),
+            self.var,
+            filepath,
+        )
         try:
             ds = xc.open_mfdataset(**args)
```
tomvothecoder, line 1015: This is the exact line where it stalls with an xarray netCDF4 lock with a NetCDF3 climatology file.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -1692,11 +1733,17 @@ def _get_default_land_sea_mask_dataset(self) -> xr.Dataset:
...
         ds_mask = xr.open_dataset(LAND_OCEAN_MASK_PATH)
-        ds_mask = squeeze_time_dim(ds_mask)
+        try:
+            ds_mask = squeeze_time_dim(ds_mask)
+            ds_mask.load(scheduler="sync")
 
-        return ds_mask
+            return ds_mask.copy(deep=True)
+        finally:
+            ds_mask.close()
```
tomvothecoder, line 2078: Load the mask dataset into memory and detach from the file before closing (resource clean up).

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -1742,17 +1798,9 @@ def _subset_vars_and_load(self, ds: xr.Dataset, var: str | list[str]) -> xr.Data
...
 
-            for dataset in (ds_subset, ds):
-                if id(dataset) in dataset_ids:
-                    continue
+        if detach:
+            return ds_subset.copy(deep=True)
 
-                dataset.close()
-                dataset_ids.add(id(dataset))
+        return ds_subset
```
tomvothecoder, line 2142: Remove closing of dataset here as it already is closed a level up in `_get_climo_dataset()` (prevent multiple `.close()` calls).

e3sm_diags/e3sm_diags_driver.py
```diff
@@ -364,8 +359,9 @@ def _run_with_dask(parameters: list[CoreParameter]) -> list[CoreParameter]:
             "again."
         )
 
-    with dask.config.set(config):
-        results = bag.map(CoreParameter._run_diag).compute(num_workers=num_workers)
+    ctx = mp.get_context("fork")
+    with ctx.Pool(processes=num_workers) as pool:
+        results = pool.map(_run_diag_with_logging, parameters, chunksize=1)
```
tomvothecoder: Use `multiprocessing.Pool` here instead of `dask.bag` because this code path is only doing a simple parallel map over independent parameter objects. `Pool.map` is a better fit for that workload: it has less scheduler overhead, fewer layers to debug, and keeps the execution model easier to reason about. In this investigation it also removed `Dask` as a possible source of the stall while preserving the same process-based parallelism and task shape.

e3sm_diags/driver/utils/dataset_xr.py
```diff
@@ -676,27 +684,103 @@ def _open_climo_dataset(self, filepath: str) -> xr.Dataset:
...
+
+        try:
+            for filepath in filepaths:
+                with netCDF4.Dataset(filepath, mode="r") as ds_climo:
+                    if ds_climo.data_model not in NETCDF3_DATA_MODELS:
+                        return False
+        except OSError:
+            return False
+
+        return True
```
tomvothecoder, line 1065: This code detects if the climatology file is a NetCDF3 file or not. If it is, then return `lock=True` for `xc.open_mfdataset()` to prevent hang.

tomvothecoder (commented):
@chengzhuzhang @xylar I made more progress on the Python 3.14 stall investigation. The version boundary is now clearer. The strongest current lead is a lock-related issue in xarray's `netcdf4` backend when opening or reading `NetCDF3` climo files.

Not sure yet whether this will be ready for E3SM Unified, but the investigation has narrowed the issue substantially.

@chengzhuzhang, could you do the zppy production run on NERSC with this branch using Python 3.14 and `xarray=2026.01.0`? I don't have it setup on NERSC and LCRC is still down. I've also made PR review comments below with additional code cleanup and optimization.

---

Current boundary:

- `xarray==2025.12.0`: no stall observed
- `xarray>=2026.01.0`: intermittent stall reproduces
- `xarray>=2026.01.0` with `lock=False` in the climo open path: 10/10 min script runs completed

The strongest current lead is the xarray `netcdf4` backend lock path during climo open/read work. In stuck workers from the reduced test script [`qa.py`](https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/parallel-nersc/min-scripts/qa.py), traces consistently land in `xarray.backends.locks`, `file_manager`, and `netCDF4_*`. In most stalled runs I inspected, one worker appears stuck waiting to acquire a lock.

The affected climo inputs are `NetCDF3` files:
- `v3.LR.historical_0051_ANN_198501_201412_climo.nc`
- `v3.LR.historical_0051_DJF_198501_201412_climo.nc`

My current read is a lock-related hang or regression in xarray's `netcdf4` backend on this Python 3.14 stack, likely exposed by the broader workflow rather than by simple parallel open/read alone. This is directionally consistent with upstream xarray history: PR [#10788](https://github.com/pydata/xarray/pull/10788) added locking around `netCDF4` close operations as a partial fix in `xarray=2026.1.0`, and issue [#824](https://github.com/pydata/xarray/issues/824) notes that read-only `NetCDF3` access should not require the same lock path (use `lock=False`).

`lock=False` now looks like a plausible workaround for this specific read-only `NetCDF3` climo path, but I have not yet validated correctness or production-scale safety.

More detail in `notes.md`:
https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/parallel-nersc/notes.md

e3sm_diags/e3sm_diags_driver.py
tomvothecoder: Reverting this to prevent scope creep in this PR.

chengzhuzhang:
@tomvothecoder thank you for a solid investigation. It sounds like an upstream issue with xarray and python3.14 when supporting netcdf3 files. I agree that we should do production runs to test this workaround, maybe several times to see if `lock=False` will cause trouble. Also not sure if having `_should_disable_climo_lock()` opens each climo file twice would have some performance implication. 

 I will test it when chrysalis is back early next Monday since my benchmark runs for this issue were also run on chrysalis.

tomvothecoder:
> @tomvothecoder thank you for a solid investigation. It sounds like an upstream issue with xarray and python3.14 when supporting netcdf3 files. I agree that we should do production runs to test this workaround, maybe several times to see if `lock=False` will cause trouble. 


I agree, we should run at least 3 times. Here are the next steps to do. Thanks for the help.
> ## Next steps
> 
> 1. Run the full production case with `lock=False` at least 3 times.
> 2. Compare metrics, JSON outputs, and saved NetCDF products against a trusted baseline, ideally `xarray==2025.12.0`.
> 3. Compare the repeated `lock=False` runs against each other for numerical stability.
> 4. If those checks pass, treat `lock=False` as a temporary workaround candidate for `xarray>=2026.01.0`.
> 5. Prepare an upstream report with the minimal version boundary:
>    - `2025.12.0` works
>    - `2026.01.0+` stalls
>    - `lock=False` removes the stall
> 6. In that report, include that xarray issue `#824` indicates netCDF3 reads should not need the lock, while this read-only netCDF3 path still appears to block in lock acquisition.
> Source:  https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/parallel-nersc/notes.md#next-steps

> Also not sure if having `_should_disable_climo_lock()` opens each climo file twice would have some performance implication.

Yes, it will be good validate for any performance difference. This function uses the `NetCDF4` package with read-only access to get the metadata and breaks out of the loop immediately if it detects any NetCDF4 file. I expect the overhead to be minimal.

chengzhuzhang:
@tomvothecoder thanks for the next steps, they are make sense to me. For the item: 
```
Compare metrics, JSON outputs, and saved NetCDF products against a trusted baseline, ideally xarray==2025.12.0.
```
As well as the performance bit, it seems streamlining this is relevant to the other issue https://github.com/E3SM-Project/e3sm_diags/pull/903

tomvothecoder:
Hey @chengzhuzhang, I ran another round of tests using Python 3.14.3 and 3.14.6 with several Xarray versions and made some promising findings. We can discuss more once you're back from vacation.

Full report: https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/test-scripts/260717-test-results/summary.md

### Key Takeaways
The results suggest that a Python change after 3.14.3 may have resolved the stall:

- **Python 3.14.3:** ❌ Both minimal and full runs stalled with Xarray >=2026.1.0.
- **Python 3.14.6:** ✅ All tested Xarray releases, bisected commits, and the full run completed without stalling.
- The environments were mostly aligned aside from Python, Xarray, and a few unrelated dependencies.
### Recommended Actions

Although a Python release newer than 3.14.3 may have resolved the stall, I recommend keeping the `lock=False` workaround for read-only NetCDF3 files. When Xarray uses the netCDF4 engine, it applies its combined NetCDF-C/HDF5 lock even though NetCDF3 files are not HDF5-backed. Disabling that lock for these read-only inputs avoids unnecessary serialization that may cause contention or deadlocks. The creator of Xarray previously recommended lock=False for read-only NetCDF3 files in this [Xarray discussion](https://github.com/pydata/xarray/issues/824#issuecomment-210633360).

Before finalizing, we should compare two full runs:

1. `main` with Python 3.13 and Xarray >=2026.1.0, without the workaround.
2. Python 3.14.3 with Xarray 2026.1.0, with the workaround.

The outputs should match, confirming that `lock=False` prevents the stall without changing diagnostic results.

### Upstream Actions
I have not been able to create a standalone reproducer suitable for the upstream Xarray repository. The stall appears to depend on E3SM Diags’ broader concurrent file-access pattern, so I do not recommend investing more time in this path.

chengzhuzhang:
@tomvothecoder thanks for the nice investigation. I will catch up when i return.

tomvothecoder:
I expanded the Python-version testing with Xarray fixed at 2026.7.0, covering Python 3.13.12–3.13.14 and 3.14.0–3.14.6, both with and without the NetCDF3 `lock=False` workaround.

Full investigation: https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/investigations/260720-xarray-lock-stall-across-py-versions/summary.md

### Results

| Python | Workaround disabled | Workaround enabled |
|---|---|---|
| 3.13.12–3.13.14 | ✅ Minimum 5/5; full runs completed | ✅ All completed |
| 3.14.0 | ✅ Minimum 5/5; full run completed | ✅ All completed |
| 3.14.1–3.14.4 | ❌ Minimum tests stalled; full runs timed out at 4 hours | ✅ All completed |
| 3.14.5–3.14.6 | ✅ Minimum 5/5; full runs completed | ✅ All completed |

### Key takeaways

- Without the workaround, the affected range is specifically **Python 3.14.1–3.14.4**. Python 3.14.0 works, and the issue disappears again in 3.14.5.
- With the workaround enabled, all 10 minimum-test jobs completed all five repetitions (50/50 total), and all 10 full diagnostic jobs completed.
- I was **not able to reproduce the stall with Python 3.13.14—or any tested Python 3.13 release**. This differs from the Python 3.13.14 `lat_lon_land` result [reported here](https://github.com/E3SM-Project/e3sm_diags/pull/1067#issuecomment-5012842006), suggesting reproducibility may also depend on the diagnostic workload or file-access pattern.
- The Python changelog points to a possible incremental-GC interaction: GC accounting changed in 3.14.1, and 3.14.5 reverted the incremental collector. This matches the observed boundaries but is not yet a proven root cause. The Codex changelog interpretation can be found in [this section](https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/investigations/260720-xarray-lock-stall-across-py-versions/summary.md#3-python-release-interpretation) of the full investigation report.

### Recommendation

Keep the `lock=False` workaround for read-only NetCDF3 climatology files. It prevents the stall across every affected Python release tested, while the underlying Python/Xarray interaction remains timing- and workload-dependent.

Across the six Python versions that completed with and without the workaround, enabled runs averaged 1 minute 28 seconds (3.2%) longer. Because the cohorts ran separately, this does not establish that the difference is workaround overhead. The small observed difference is outweighed by the workaround allowing affected Python 3.14.1–3.14.4 runs to complete in approximately 47 minutes instead of timing out after four hours.

### Output equivalence validation

I compared paired full-run outputs for Python 3.13.14 and 3.14.6, with the workaround disabled versus enabled.

Validation report: https://github.com/E3SM-Project/e3sm_diags/blob/bugfix/py314-stall-cont/auxiliary_tools/debug/1048-py314-stall-cont/investigations/260720-xarray-lock-stall-across-py-versions/validation/netcdf3-climo-lock-workaround-20260721/summary.md

| Python | Inventory | NetCDF exact | JSON canonical exact | PNG pixel exact | Max abs/rel difference | Result |
|---|---:|---:|---:|---:|---:|---|
| 3.13.14 | 3,300 common; 0 missing/additional | 5/5 | 1,428/1,428 | 1,867/1,867 | 0 / 0 | ✅ PASS |
| 3.14.6 | 3,300 common; 0 missing/additional | 5/5 | 1,428/1,428 | 1,867/1,867 | 0 / 0 | ✅ PASS |

Validation details:

- The NetCDF comparisons covered structure, dimensions, variables, metadata, dtypes, NaN masks, exact values, and `atol=1e-12` / `rtol=1e-12` tolerance status.
- The canonical JSON comparisons were exact. The only raw file mismatch was `viewer/index.json`, due to three embedded results-directory strings and one generation timestamp; all four expected provenance differences disappear after canonicalization.
- Every PNG was byte-identical and was also decoded to RGBA for an independent pixel comparison. All 1,867 images per pair were pixel-identical, with zero differing pixels, zero maximum channel difference, and zero RMS difference.
- The incomplete Python 3.14.1–3.14.4 disabled outputs were excluded from equivalence validation.

tomvothecoder:
@chengzhuzhang I added another [investigation report and output validation](https://github.com/E3SM-Project/e3sm_diags/pull/1048#issuecomment-5039876089) for your review above.

The stall appears to involve an interaction between Xarray’s lock behavior introduced in `xarray>=2026.01.0` ([PR #10788](https://github.com/pydata/xarray/pull/10788)) and Python 3.14 garbage-collection changes. The affected range, Python 3.14.1–3.14.4, aligns with incremental-GC changes that were reverted in Python 3.14.5, although this is not yet a proven root cause.

Using `lock=False` for read-only NetCDF3 climatology files remains the best path forward: it prevents the stall, has minimal observed performance impact, and produces identical NetCDF, JSON, and PNG outputs in the completed validation runs.

The next step is to test the workaround with the Python 3.13.14 `lat_lon_land` case from your [PR comment](https://github.com/E3SM-Project/e3sm_diags/pull/1067#issuecomment-5012842006), since I could not reproduce the stall with any Python 3.13 version in the ATM workflow.

chengzhuzhang:
Validation data point from Chrysalis (LCRC), in case it's useful here.

This fix also appears to resolve #1066, which had been attributed to the `fork` multiprocessing context. That repro ran on e3sm-unified 1.13.0 (xarray 2026.2.0), inside the regression range this PR targets. Same data, `fork`, no guard:

| code | context | xarray | result |
|---|---|---|---|
| main | fork | 2026.7.0 | hung, 1632 plots |
| forkserver | forkserver | 2026.7.0 | hung, 1667 plots |
| main | fork | 2025.12.0 | OK, 1697 plots |
| **main + this PR** | **fork** | **2026.7.0** | **OK, 1697 plots** |

Applying this PR gives the same result as downgrading xarray below the break — 1697 plots either way.

Details: `lock=False` fired 5390× on 0.25° NetCDF3 (CDF-5) ELM land climo files (1440×720, ~2.4 GB each). No runtime cost — 1:14:02 vs a 1:16:02 baseline. This is a **python 3.13.14 + xarray 2026.7.0 + `lat_lon_land`** cell, i.e. py3.13 with a land set.

chengzhuzhang (approved):
@tomvothecoder thanks for the heroic troubleshooting and testing. I think this fix can advance e3sm_diags to python 3.14 support. I also verified with a test that resolves #1066 . Could you update the PR title or description, maybe also do a PR clean up before merging?

tomvothecoder:
> @tomvothecoder thanks for the heroic troubleshooting and testing. I think this fix can advance e3sm_diags to python 3.14 support. I also verified with a test that resolves #1066 . Could you update the PR title or description, maybe also do a PR clean up before merging?

@chengzhuzhang That's great to hear that `lock=False` works on your branch too! Thank you for testing.

Yes, I will update the PR title and description. I will clean the branch to keep only the the necessary scripts and artifacts, and generate a concise final report in the respective `debug` folder.

PR #1059, opened 2026-05-27
Handle missing AEW histogram file in tc_analysis diagnostic

### LINKED ISSUE #1058: tc_analysis_driver crashes when AEW histogram file is missing
## Summary

`tc_analysis_driver.run_diag()` unconditionally opens the AEW histogram file (`aew_hist_*.nc`) at line 85, causing a `FileNotFoundError` when the file does not exist. The upstream zppy `tc_analysis.bash` template already handles this gracefully — it skips AEW histogram generation when no AEW tracks are found — but e3sm_diags does not account for the file being absent.

When the AEW file is missing, the entire `tc_analysis` diagnostic fails, even though the cyclone track density, TC intensity, TC frequency, ACE distribution, and seasonal cycle plots could still be produced.

## Error

```
FileNotFoundError: [Errno 2] No such file or directory:
  '.../tc-analysis_1995_2004/aew_hist_<caseid>_1995_2004.nc'
```

Traceback points to `tc_analysis_driver.py:85`:
```python
test_aew_hist = xr.open_dataset(test_aew_file).sel()["density"]
```

## Suggested Fix

**`e3sm_diags/driver/tc_analysis_driver.py`**:
1. Check `os.path.exists(test_aew_file)` before opening; log a warning and set to `None` if missing.
2. Only populate `test_data["aew_density"]` / `test_data["aew_num_years"]` when AEW data exists.
3. Apply the same guard for the `model_vs_model` ref AEW path (line ~124).

**`e3sm_diags/plot/tc_analysis_plot.py`**:
1. Guard the `plot_map(test, ref, "aew", parameter)` call (line ~221) — only plot the AEW density map when both `test` and `ref` contain `"aew_density"`.

## Reproduction

Run zppy `tc_analysis` on an EAMxx case where no African Easterly Waves are detected (e.g., short simulation or configuration without sufficient 850 hPa wind variability), then run `e3sm_diags` with `sets = "tc_analysis"`. The AEW histogram file will not be generated by zppy, and e3sm_diags will crash.

chengzhuzhang:
## Summary

Fixes #1058.

- Guard all three AEW histogram file loads (test, model_vs_model ref, model_vs_obs ref) with `os.path.exists()` in `tc_analysis_driver.py`
- Skip the AEW density plot in `tc_analysis_plot.py` when either test or ref AEW data is absent
- Log a warning when AEW file is missing instead of crashing the entire diagnostic

The upstream zppy `tc_analysis.bash` template already gracefully skips AEW histogram generation when no AEW tracks are found, but e3sm_diags did not account for the file being absent — causing a `FileNotFoundError` that blocked all other TC plots (intensity, frequency, ACE, seasonal cycle, cyclone density) from being produced.

## Test plan

- [x] Run `tc_analysis` set on a case where AEW histogram file exists — all plots (including AEW density) should be generated as before
- [x] Run `tc_analysis` set on a case where AEW histogram file is missing — remaining TC plots should succeed, with a warning logged about the missing AEW file

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/e3sm_diags/driver/tc_analysis_driver.py
+++ b/e3sm_diags/driver/tc_analysis_driver.py
@@ -82,7 +82,6 @@ def run_diag(parameter: TCAnalysisParameter) -> TCAnalysisParameter:
         test_data_path,
         "aew_hist_{}_{}_{}.nc".format(test_name, test_start_yr, test_end_yr),
     )
-    test_aew_hist = xr.open_dataset(test_aew_file).sel()["density"]
 
     test_data = collections.OrderedDict()
     ref_data = collections.OrderedDict()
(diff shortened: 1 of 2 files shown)


chengzhuzhang:
A new run was tested where African Eastern Wave (AEW) data is missing, producing expected[ results](https://portal.nersc.gov/cfs/e3sm/chengzhu/tests/zppy_for_eamxx_05262026/ne256pg2_ne256pg2.F20TR-SCREAMv1.July-1.spanc800.2xauto.acc150.n0032.test2.1/e3sm_diags/atm_monthly_180x360_aave/model_vs_obs_1995-2004/viewer/tc_analysis/index.html).

PR #1075, opened 2026-08-10
Pin `proj <9.8` and refresh Python 3.14 image baselines

tomvothecoder:
## Summary

- Pin native conda `proj <9.8` in CI and development environments to work around Cartopy <0.26 `PlateCarree` compatibility regression.
- Refresh polar image baselines using pinned `proj` and make Python 3.14 the baseline-authority environment.
- Record `dask`, native `proj`, and `pyproj` versions in baseline metadata.

The pin is intentionally temporary. It avoids `proj` 9.8 rendering/clipping regressions while a durable Cartopy/proj-compatible solution is evaluated.

Related: #1077

## Validation

- `pytest tests/integration/test_plot_image_regressions.py tests/integration/test_refresh_plot_image_baselines.py` — 10 passed on Python 3.14.6
- Pre-commit hooks passed (including Ruff and mypy)

## Notes

The native `proj` constraint remains conda-only; `pyproject.toml` does not install unrelated PyPI `proj` package.


tomvothecoder:
## CI failure: Python 3.14 broad integration

The Python 3.14 broad-integration job fails while Cartopy saves a global `lat_lon` plot:

`ValueError: 'vertices' must be 2D with shape (N, 2), but your input has shape (0,)` ([example](https://github.com/E3SM-Project/e3sm_diags/actions/runs/31423051087/job/93569622617#step:12:184))

The error originates in Cartopy’s GeoAxes projection/clipping path. Other current matrix jobs were cancelled by fail-fast; they were not independent failures.

| Package | [Aug. 3 pass — Python 3.14](https://github.com/E3SM-Project/e3sm_diags/actions/runs/30858439650/job/91834805777) | [Aug. 10 fail — Python 3.14](https://github.com/E3SM-Project/e3sm_diags/actions/runs/31423051087/job/93569622617) |
|---|---:|---:|
| Python | 3.14.6 | 3.14.6 |
| Cartopy | 0.24.0 | 0.24.0 |
| Matplotlib | 3.11.1 | 3.11.1 |
| NumPy | 2.4.6 | 2.4.6 |
| Shapely | 2.1.2 | 2.1.2 |
| GEOS | 3.14.1 | 3.14.1 |
| **PROJ** | **9.7.1** | **9.8.1** |
| pyproj | 3.7.2, build 3 | 3.7.2, build 5 |
| GDAL | 3.12.3 | 3.13.2 |
| rasterio | 1.5.0 | 1.5.1 |
| pyogrio | 0.12.1 | 0.13.0 |

Both solves contain 364 packages; 47 package versions/builds changed, with no additions or removals. The dependency specifications were unchanged, so this was an upstream conda-forge resolution change.

## Root cause (GPT Analysis, unverified)

PROJ 9.8 changed the Equidistant Cylindrical (`+proj=eqc` / Plate Carrée) projection to add ellipsoidal calculations: https://github.com/OSGeo/PROJ/pull/4656. [PROJ 9.8 release notes](https://proj.org/en/stable/news.html)

Cartopy 0.24's `PlateCarree` creates an ellipsoidal WGS84 `+proj=eqc` CRS. The failing global `lat_lon` plot uses `PlateCarree(central_longitude=180)`, which now takes PROJ 9.8's new ellipsoidal `eqc` path. This breaks Cartopy's clipping boundary and produces the empty-vertices Matplotlib error.

This is not explained by PROJ's EPSG database update: PROJ 9.8.1 reverted that database to the same version used by PROJ 9.7.1, but the failure remained.


## Workaround 1 - Temporarily pin `proj<9.8` (verified)
[PR #1075](https://github.com/E3SM-Project/e3sm_diags/pull/1075) constrains `proj <9.8` and passes the full Python matrix. This confirms the issue is not Python 3.14-specific.

`proj <9.8` is an appropriate temporary compatibility constraint until Cartopy is upgraded or an upstream Cartopy/PROJ fix is available.

## Workaround 2 - Change construction of `PlateCarree` (unverified)

Rather than pinning PROJ, E3SM Diags could explicitly construct a **spherical** Cartopy `PlateCarree` globe. This should reproduce PROJ 9.7's effective `eqc` behavior and avoid PROJ 9.8's new ellipsoidal path.

```python
SPHERICAL_GLOBE = ccrs.Globe(
    ellipse=None,
    semimajor_axis=6_378_137.0,
    semiminor_axis=6_378_137.0,
)

def plate_carree(central_longitude: float = 0) -> ccrs.PlateCarree:
    return ccrs.PlateCarree(
        central_longitude=central_longitude,
        globe=SPHERICAL_GLOBE,
    )
```

The initial test would replace the `ccrs.PlateCarree()` calls in `lat_lon_plot.py`, including the global `central_longitude=180` axes projection and its source `transform` CRS, with this helper. If it fixes the failing broad integration test with `proj >=9.8`, expand it to other E3SM Diags plotting modules that use `PlateCarree`.

This needs image-regression coverage before adoption, but it is a possible local compatibility fix that would allow removal of the `proj <9.8` constraint without waiting for an upstream Cartopy or PROJ release.

tomvothecoder:
@chengzhuzhang The integration tests began failing recently. My https://github.com/E3SM-Project/e3sm_diags/pull/1075#issuecomment-5245489829 details the investigation.

The likely regression is PROJ 9.8’s new ellipsoidal Equidistant Cylindrical / Plate Carrée (`eqc`) support: https://github.com/OSGeo/PROJ/pull/4656. Cartopy 0.24 uses this projection for E3SM Diags’ global `lat_lon` plots.

Pinning `proj <9.8` works around the issue and allows the full CI build to pass.

A possible alternative is to update E3SM Diags to explicitly use a spherical Cartopy `PlateCarree` globe. That should preserve the effective PROJ 9.7 behavior and avoid PROJ 9.8’s new ellipsoidal `eqc` path. This would need broad-integration and image-regression testing before replacing the dependency pin.

xylar:
@tomvothecoder and @chengzhuzhang, this constraint will have to be temporary. We cannot have it for the next unified. I can almost guarantee that it will make the metapackage unsolvable.

tomvothecoder:
> @tomvothecoder and @chengzhuzhang, this constraint will have to be temporary. We cannot have it for the next unified. I can almost guarantee that it will make the metapackage unsolvable.

@xylar This is what I had in mind too. 

I think this issue stems from a Cartopy + proj>=9.8 compatibility issue that affects existing Cartopy releases. There is an open PR that is meant to fix this in Cartopy: https://github.com/SciTools/cartopy/pull/2653. I believe any package in E3SM Unified that uses Cartopy's `PlateCarree` _might_ be affected by `proj>=9.8` changes. More context can be found in this Cartopy v0.26 PR too: https://github.com/SciTools/cartopy/pull/2641#issuecomment-5225438446

We can try to implement a code workaround on the `e3sm_diags` side, but it'll probably be pretty invasive and doesn't guarantee that other packages in E3SM Unified won't break too. 

I think `e3sm_diags` only started picking up `proj >=9.8` this past week when another dependency updated its dependency constraints. Before `e3sm_diags` was only picking up `proj <=9.7.1`, which worked fine.

tomvothecoder:
> We can try to implement a code workaround on the `e3sm_diags` side, but it'll probably be pretty invasive and doesn't guarantee that other packages in E3SM Unified won't break too.

I'm using Codex to try to implement workaround 2 in case Cartopy v0.26 is not released in time with the upstream fix.
We can keep that PR open in case we go with that route for E3SM Unified.

chengzhuzhang:
I agree we can pin proj as a temporal solution. It looks like the proj-9.8 update has broad impact on downstream use of `cartopy`, and hopefully there will be a fix or at least a clear direction soon.

tomvothecoder:
I merged this PR to constrain `proj <9.8` temporarily. I opened #1078 to implement workaround 2, which should hopefully allow us to remove the constraint.

PR #1090, opened 2026-09-18
Fix MERRA2 land blocks that overwrote the global TREF plots

chengzhuzhang:
## Description

In `lat_lon_model_vs_obs.cfg`, three MERRA2 blocks (`TREFMNAV`, `TREFMXAV`, `TREF_range`) are meant to be **land** plots with bilinear regridding. Each sets `regions = ["land"]` and then, a few lines later, `regions = ["global"]`:

```
variables = ["TREF_range"]
regions = ["land"]
ref_name = "MERRA2"
reference_name = "MERRA2 Reanalysis"
regions = ["global"]      <- the later value wins
...
regrid_method = "bilinear"
```

The later value wins, so each became a second **global** block. It wrote the same `MERRA2-<var>-<season>-global.{png,json}` as the real global block, which uses the default regridding. The two run in parallel and whichever finishes last overwrites the other. The global plots and metrics therefore alternate between two results from run to run, and the land plots were never produced.

### How it showed up

zppy's weekly complete run kept flagging `MERRA2-TREF_range-SON-global` as changed, and then unchanged, with the same code and data:

- The run log records `Metrics saved in …/MERRA2-TREF_range-SON-global.json` twice in every run.
- Only the `ref_regrid.*` and `diff.*` metrics change. The `test.*` (model) metrics never do, which is consistent with two regridding methods applied to one fixed reference file (`MERRA2_SON_198009_201611_climo.nc`).
- Runs land in one of exactly two states (e.g. SON ref max 21.275 vs 20.858, RMSE 1.502 vs 1.439). This happened across nearly identical environments: the 2026-09-15 and 2026-09-16 runs differ only in `threadpoolctl` and `freexl`, yet land in different states.
- Any season can flip; JJA did too. The E3SM-Unified 1.13 results are in one of the two states, so released versions are affected.

### Change

- Remove the stray `regions = ["global"]` from the three land blocks, so they produce land plots again.
- Remove the redundant duplicate `regions = ["global"]` from the three matching global blocks.
- Add `tests/e3sm_diags/test_default_diags_cfgs.py`, which checks that no two blocks in `lat_lon_model_vs_obs.cfg` write the same output. It fails on `main` and passes with this change.

Parsed with `CoreParser`, the 51 MERRA2 global outputs for these variables (3 variables × 17 seasons) were each written twice on `main`. With this change, 0 of the 102 outputs collide: the 51 global ones plus the 51 restored land ones.

### Expected effect on baselines

- The MERRA2 `TREFMNAV`, `TREFMXAV` and `TREF_range` global results become deterministic, using the global block's default regridding. Any baseline that happened to capture the bilinear variant will show a one-time difference.
- New `MERRA2-{TREFMNAV,TREFMXAV,TREF_range}-<season>-land` plots appear.

### Follow-ups (not in this PR)

A scan of `default_diags/` for keys repeated within one block found three more blocks that look like two blocks merged by a missing `[#]` separator. In each, the second half's values silently replace the first's:

- `lat_lon_model_vs_obs.cfg` (`aero-no-ref-data`) and `lat_lon_model_vs_model.cfg`: `_Mass_soa_srf` is replaced by `_Mass_mom_srf`, so `_Mass_soa_srf` is never plotted.
- `zonal_mean_xy_model_vs_obs.cfg` (ERA5) and `zonal_mean_xy_model_vs_model.cfg`: the `U` 850/200 hPa plots are replaced by `Z3` 500 hPa.

Fixing those adds plots rather than removing a race, so they are left for a separate change.

## Checklist

If applicable:

  - The new test passes. `tests/e3sm_diags/test_parameters.py::test_area_mean_time_series_parameter` fails locally both with and without this change (a log-capture assertion; this change touches no parser or parameter code).
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have commented my code, particularly in hard-to-understand areas

🤖 Generated with [Claude Code](https://claude.com/claude-code)


chengzhuzhang:
@tomvothecoder  FYI, this is a fix in a configuration file. I’ll merge it, and it should address the zppy weekly test diff.

